PR #6, opened 2022-06-10
Update to vscode api 1.68

### LINKED ISSUE #2: Possible issues with VSCode->Monaco treeshaking
Vscode treeshaking is really aggresive. This lib tries to use as much monaco code as possible instead of bundling vscode code.

The problem is that some monaco class will be used, and sometime, methods were treeshaked by the VSCode->Monaco build process.

Examples:
- `IMarkerService.changeAll`
- `VSBuffer.alloc`

The current build process is unable to detect it.

Possible solutions:
- Detect that a monaco class was partially treeshaked and use the vscode class instead
- Detect that a monaco class was partially treeshaked and extend it to add missing methods/stuff

### LINKED ISSUE #3: Treeshaking not optimal
The rollup treeshaking is unable to treeshake some unused parts of the code
- Some classes are unused but not treeshaked (example: `FileOperationError`) (a second rollup run would success to treeshake it though)
- Methods that are not used (especially in `MainThreadLanguageFeatures`, only static methods are used but all unused methods are kept)

CGNonofr:
fix #3 
partially fix #2 (But I don't think there is any 100% robust solution)


CGNonofr:
Do not forget to export the new classes

PR #26, opened 2022-07-19
Add more apis

CGNonofr:


src/service-override/messages.ts
```diff
@@ -20,7 +20,7 @@ import { unsupported } from '../tools'
 class LayoutService implements ILayoutService, Pick<IWorkbenchLayoutService, 'isVisible'> {
   declare readonly _serviceBrand: undefined
 
-  constructor (public container: HTMLElement, private _codeEditorService: ICodeEditorService) {
+  constructor (public container: HTMLElement, @ICodeEditorService private _codeEditorService: ICodeEditorService) {
```
BeLi4L, line 23: What does that do?

CGNonofr: It is used by the vscode/monaco injection mechanism

CGNonofr: (So it was undefined without the annotation)

BeLi4L (approved):

PR #32, opened 2022-07-21
fix: build for darwin

dlvhdr:


dlvhdr:
@CGNonofr I'm not sure if the `sed` command works like this in linux.
Regarding `rsync` I pretty sure linux has it installed.

CGNonofr:
Thanks! actually, you can even remove the sed and it's not useful anymore!

dlvhdr:
@CGNonofr Done!

CGNonofr:
I fixed the github action trigger, can you please rebase on main to check the build in the github action?

CGNonofr:
Ok it seems to work! One last thing, can you please follow the [conventionnal commit](https://www.conventionalcommits.org/en/v1.0.0/#summary) convention?

CGNonofr:
> Something like this? Not sure if I followed it correctly

Perfect! I would have probably labelled the sed removal by a `clean` instead of a `fix` but it's no big deal

CGNonofr (approved):

PR #56, opened 2022-11-28
Declare monaco-editor as peerDependency

kaisalmen:
As discussed in https://github.com/TypeFox/monaco-languageclient/pull/435 `monaco-editor` is added as peerDependency. Additionally the range for both the dependency and the peerDependency is limited to >= 0.34.0 < 0.35.0 via "~".


kaisalmen:
One afterthought: There is use case for defining both: When you want to enforce a "wider" range as peer (with dependent projects in mind) and a specific dependency. This is not applicable here (currently), but there are scenarios where it makes sense.

PR #69, opened 2023-02-27
Demo and Build improvements

kaisalmen:
List of enhancements:

- Update demo specific dependencies
- Demo: change `monaco-vscode-api` dependency to local reference.
- `npm install` will no longer pull vscode if version info file with same version content is found. It is checked before the download starts and the file is written at the end
- vscode+eslint works now for alle code. `tsconfig.eslint.json` + eslint config changes are introduced + eslint related settings for vscode are added.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,6 +3,12 @@
     "es6": true,
     "browser": true
   },
+  "parser": "@typescript-eslint/parser",
+  "parserOptions": {
+    "ecmaVersion": 2020,
+    "sourceType": "module",
+    "project": "./tsconfig.eslint.json"
+},
   "globals": {
     "Thenable": "readonly",
     "VSCODE_VERSION": "readonly"
@@ -10,7 +16,13 @@
   "extends": [
     "@codingame"
   ],
-  "ignorePatterns": "vscode/**/*",
-  "plugins": ["@typescript-eslint", "import", "unused-imports"]
-
+  "ignorePatterns": [
+    "vscode/**/*",
+    "dist/**/*"
+  ],
+  "plugins": [
+    "@typescript-eslint",
+    "import",
+    "unused-imports"
+  ]
 }
--- a/demo/package.json
+++ b/demo/package.json
@@ -4,21 +4,24 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "start": "vite"
+    "clean": "tsc -b -c",
+    "compile": "tsc",
+    "start": "vite",
+    "start:debug": "vite --debug --force"
   },
   "devDependencies": {
-    "@types/throttle-debounce": "^5.0.0",
-    "typescript": "^4.7.4",
-    "vite": "^3.0.9"
+    "@types/throttle-debounce": "~5.0.0",
+    "typescript": "~4.9.5",
+    "vite": "~4.1.4"
   },
   "dependencies": {
-    "monaco-editor": "~0.34.0",
-    "throttle-debounce": "^5.0.0",
-    "vscode": "npm:@codingame/monaco-vscode-api@~1.69.21",
+    "monaco-editor": "~0.34.1",
+    "throttle-debounce": "~5.0.0",
+    "vscode": "file:../",
     "vscode-oniguruma": "~1.6.2"
   },
   "volta": {
-    "node": "16.15.0",
-    "npm": "8.11.0"
+    "node": "18.14.2",
+    "npm": "9.5.1"
   }
 }
--- a/demo/src/main.ts
+++ b/demo/src/main.ts
@@ -13,16 +13,15 @@ import 'monaco-editor/esm/vs/editor/standalone/browser/referenceSearch/standalon
 import 'monaco-editor/esm/vs/language/json/monaco.contribution'
 import { getUserConfiguration, onUserConfigurationChange, updateUserConfiguration } from 'vscode/service-override/configuration'
 import { updateUserKeybindings } from 'vscode/service-override/keybindings'
-import { createConfiguredEditor } from 'vscode/monaco'
-import { getJsonSchemas, onDidChangeJsonSchema } from 'vscode/monaco'
+import { createConfiguredEditor, getJsonSchemas, onDidChangeJsonSchema } from 'vscode/monaco'
 import { debounce } from 'throttle-debounce'
 import * as vscode from 'vscode'
 
-vscode.window.showInformationMessage('Hello', {
+void vscode.window.showInformationMessage('Hello', {
   detail: 'Welcome to the monaco-vscode-api demo',
   modal: true
 }).then(() => {
-  vscode.window.showInformationMessage('Try to change the settings or the configuration, the changes will be applied to all 3 editors')
+  void vscode.window.showInformationMessage('Try to change the settings or the configuration, the changes will be applied to all 3 editors')
 })
 
 createConfiguredEditor(document.getElementById('editor')!, {
@@ -55,13 +54,13 @@ const settingsModel = monaco.editor.createModel(
 createConfiguredEditor(document.getElementById('settings-editor')!, {
   model: settingsModel
 })
-settingsModel.onDidChangeContent(debounce(1000, () => {
-  updateUserConfiguration(settingsModel.getValue())
+settingsModel.onDidChangeContent(debounce(1000, async () => {
+  await updateUserConfiguration(settingsModel.getValue())
 }))
-updateUserConfiguration(settingsModel.getValue())
+void updateUserConfiguration(settingsModel.getValue())
 onUserConfigurationChange(async () => {
   const newConfiguration = await getUserConfiguration()
-  if (newConfiguration != settingsModel.getValue()) {
+  if (newConfiguration !== settingsModel.getValue()) {
     settingsModel.setValue(newConfiguration)
   }
 })
@@ -82,10 +81,10 @@ const keybidingsModel = monaco.editor.createModel(
 createConfiguredEditor(document.getElementById('keybindings-editor')!, {
   model: keybidingsModel
 })
-keybidingsModel.onDidChangeContent(debounce(1000, () => {
-  updateUserKeybindings(keybidingsModel.getValue())
+keybidingsModel.onDidChangeContent(debounce(1000, async () => {
+  await updateUserKeybindings(keybidingsModel.getValue())
 }))
-updateUserKeybindings(keybidingsModel.getValue())
+void updateUserKeybindings(keybidingsModel.getValue())
 
 function updateDiagnosticsOptions () {
   monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
--- a/package.json
+++ b/package.json
@@ -135,11 +135,12 @@
     "README.md"
   ],
   "scripts": {
-    "build": "rm -rf dist/ && npm run lint && npm run compile && npm run generate-types && npm run compile-webpack-loader",
+    "build": "npm run clean && npm run compile && npm run lint && npm run generate-types && npm run compile-webpack-loader",
     "compile": "rollup --config rollup/rollup.config.ts --configPlugin 'typescript={tsconfig: `tsconfig.rollup.json`}' --vscode-version ${npm_package_config_vscode_version}",
+    "clean": "rm -rf dist/",    
     "compile-webpack-loader": "tsc --target es2020 -lib es2020 --module commonjs --skipLibCheck --moduleResolution node --declaration --outDir dist src/webpack-loader.ts && mv dist/webpack-loader.js dist/webpack-loader.cjs",
     "preprepare": "./scripts/install-vscode",
-    "lint": "eslint --ext ts src",
+    "lint": "eslint {**/src/**/*.ts,./rollup/*.ts,./*.ts}",
     "generate-types": "tsc --project tsconfig.types.json && rollup --config rollup/rollup.types.config.ts --configPlugin 'typescript={tsconfig: `tsconfig.rollup.json`}' && rm -rf ./dist/types",
     "release": "node --loader ts-node/esm release.ts ${npm_package_config_vscode_version}"
   },
@@ -196,7 +197,7 @@
     }
   },
   "volta": {
-    "node": "18.13.0",
-    "npm": "9.4.1"
+    "node": "18.14.2",
+    "npm": "9.5.1"
   }
 }
--- a/scripts/install-vscode
+++ b/scripts/install-vscode
@@ -5,6 +5,15 @@ vscodeVersion=$(node -e "console.log(require('./node_modules/monaco-editor/packa
 
 output_directory="`pwd`/vscode"
 patch_file="`pwd`/scripts/vscode.patch"
+version_info=$output_directory/version.info
+
+if [[ -e $version_info && $(cat $version_info) == $vscodeVersion ]]; then
+    echo "vscode version $vscodeVersion is already installed. Aborting..."
+    exit 0
+else
+    echo "vscode version $vscodeVersion is not available. Installing..."
+fi
+
 rm -rf $output_directory
 
 build_directory=`mktemp -d`
@@ -24,6 +33,9 @@ cd src
 rm -rf `find . -name '*.test.ts' -o -name 'test' -o -name 'electron-browser'`
 rm -rf vs/code
 rm -rf vs/server
+
+mkdir -p $output_directory
+echo $vscodeVersion > $version_info
 
 echo "Building vscode..."
 npx tsc --declaration --importHelpers --module es2020 --preserveConstEnums --outDir "$output_directory/vs"
--- a/tsconfig.eslint.json
+++ b/tsconfig.eslint.json
@@ -1,0 +1,15 @@
+{
+	"extends": "./tsconfig.json",
+	"compilerOptions": {
+		"noEmit": true,
+	},
+	"include": [
+		"**/src/**/*.ts",
+		"./rollup/*.ts",
+		"./*.ts"
+	],
+	"exclude": [
+		"./dist/**/*",
+		"./vscode/**/*"
+	]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,14 @@
+{
+  "eslint.codeAction.showDocumentation": {
+    "enable": true
+  },
+  "eslint.validate": [
+    "javascript",
+    "typescript"
+  ],
+  "eslint.workingDirectories": [
+    "**/src/**/*.ts",
+    "./rollup/*.ts",
+    "./*.ts"
+  ]
+}
--- a/README.md
+++ b/README.md
@@ -10,8 +10,8 @@ The VSCode api is composed of:
   - If it's an important feature: we let the user implement it as they wish.
   - If it's some advanced features that don't make a lot of sense on Monaco (debug, tests...), it just throws an error when you try to use it.
 
-
 To implement by hands the optional features (file system, workspace folders, file...), you can use the `Services` namespace from `vscode/services`:
+
 ```typescript
 import { Services } from 'vscode/services'
 Services.install({
@@ -37,6 +37,7 @@ This library allows you to use a more convenient way using `StandaloneServices.i
 Also, monaco-editor doesn't provide good type for them, so this library does it.
 
 Example:
+
 ```typescript
 import { StandaloneServices, INotificationService } from 'vscode/services'
 
@@ -49,6 +50,7 @@ StandaloneServices.initialize({
 ```
 
 Additionally, this library exposes 10 modules that include the vscode version of some services (with some glue to make it work with monaco):
+
 - Notifications: `vscode/service-override/notifications`
 - Dialogs: `vscode/service-override/dialogs`
 - Model / Editor: `vscode/service-override/modelEditor`
@@ -62,6 +64,7 @@ Additionally, this library exposes 10 modules that include the vscode version of
 - Token classification: `vscode/service-override/tokenClassification`
 
 Usage:
+
 ```typescript
 import { StandaloneServices } from 'vscode/services'
 import getModelEditorServiceOverride from 'vscode/service-override/modelEditor'
@@ -106,12 +109,14 @@ The editors created using `monaco.editor.create` don't use the configuration fro
 This library exposes functions to create editors binded on the configuration:
 
 before:
+
 ```typescript
 import * as monaco from 'monaco-editor'
 monaco.editor.create(...)
 ```
 
 after:
+
 ```typescript
 import { createConfiguredEditor } from 'vscode/monaco'
 
@@ -143,6 +148,7 @@ vscode.languages.registerCompletionItemProvider(...)
 
 There is a demo that showcases the service-override features. It allows to register contributions with the same syntaxes as in VSCode.
 It includes:
+
 - Languages
 - Language configurations
 - VSCode themes
@@ -157,10 +163,13 @@ It includes:
 It also uses the `getJsonSchemas` function to register them on the monaco json worker and have autocomplete/hover on settings and keybindings.
 
 From CLI run:
+
 ```bash
 cd demo
 npm ci
-npm start
+npm run start
+# OR: for vite debug output
+npm run start:debug
 ```
 
 ### History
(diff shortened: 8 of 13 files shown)


demo/package.json
```diff
@@ -4,21 +4,20 @@
...
-    "vite": "^3.0.9"
+    "@types/throttle-debounce": "~5.0.0"
   },
   "dependencies": {
-    "monaco-editor": "~0.34.0",
-    "throttle-debounce": "^5.0.0",
-    "vscode": "npm:@codingame/monaco-vscode-api@~1.69.7",
-    "vscode-oniguruma": "~1.6.2"
+    "throttle-debounce": "~5.0.0",
+    "vscode": "npm:@codingame/monaco-vscode-api@~1.69.21",
```
CGNonofr, line 16: Shouldn't we use a wildcard here?

package.json
```diff
@@ -137,11 +137,11 @@
...
     "compile": "npm run compile:local --workspaces",
-    "build": "rm -rf dist/ && npm run lint && npm run compile:local && npm run generate-types && npm run compile-webpack-loader",
-    "clean:local": "echo Nothing to clean",
-    "compile:local": "rollup --config rollup/rollup.config.ts --configPlugin 'typescript={tsconfig: `tsconfig.rollup.json`}' --vscode-version ${npm_package_config_vscode_version}",
+    "build": "npm run clean:local && npm run compile:local && npm run lint && npm run generate-types && npm run compile-webpack-loader",
+    "clean:local": "rm -rf dist/",
+    "compile:local": "npm run prepare:vscode && rollup --config rollup/rollup.config.ts --configPlugin 'typescript={tsconfig: `tsconfig.rollup.json`}' --vscode-version ${npm_package_config_vscode_version}",
     "compile-webpack-loader": "tsc --target es2020 -lib es2020 --module commonjs --skipLibCheck --moduleResolution node --declaration --outDir dist src/webpack-loader.ts && mv dist/webpack-loader.js dist/webpack-loader.cjs",
-    "preprepare": "./scripts/install-vscode",
+    "prepare:vscode": "./scripts/install-vscode",
```
CGNonofr, line 144: Why do we need this change?

kaisalmen: It bugs me that `npm i` always run the vscode installation, but my proposed solution is not good. I will come up with a better idea tomorrow.

Do you think it is a good idea to have one workspace being a sub-directory of another workspace (contained in I mean)? So far, I only have created workspace that don't overlap.

CGNonofr: > I will come up with a better idea tomorrow.

I agree a kind a cache could be nice! we can probably skip the script if the vscodeRef didn't change

> Do you think it is a good idea to have one workspace being a sub-directory of another workspace (contained in I mean)? So far, I only have created workspace that don't overlap.

I won't say I'm an expert of npm workspaces, they are a very hard thing to understand and master and that was the reason I didn't use it at first so I don't really know 🤷

kaisalmen: > I agree a kind a cache could be nice! we can probably skip the script if the vscodeRef didn't change

I have build this in now ⬇️ 

> I won't say I'm an expert of npm workspaces, they are a very hard thing to understand and master and that was the reason I didn't use it at first so I don't really know 🤷

It works for now. Longer term, I would suggest to move the main package into its own sub dir like:
```shell
api
demo
```

kaisalmen:
I have updated the main description!

package.json
CGNonofr: It bothers me a little to have to move the main package into a subdirectory just to be able to add a little demo project next to it

kaisalmen: yes, that's why I haven't done it, so far. Maybe it is a better idea to exclude it from the package and use vite and separate tsconfig for the demo and move it to the main source folder. WDYT?

kaisalmen: Should I build a POC on a separate commit?

CGNonofr: It looks like a good idea!

kaisalmen:
@CGNonofr I have moved the demo code to `src/demo`. It excluded from the rollup build via exclusion in `tsconfig.build.json`

Code is still considered by vscode / eslint because it is contained in `tsconfig.json`

vite requires the project to be build because in `package.json` use the project itself as `devDependency`
```json
 "vscode": "."
```

CGNonofr:
> Build is green now. vite resolves the packages, so there is no need to self-reference in the `package.json` now. Please check if the demo works as expected.

I have the `ProductQualityContext.notEqualsTo` error, did you rebase on main?

also the README needs to be updated

src/demo/main.ts
```diff
@@ -13,16 +13,15 @@ import 'monaco-editor/esm/vs/editor/standalone/browser/referenceSearch/standalon
...
 import { getUserConfiguration, onUserConfigurationChange, updateUserConfiguration } from 'vscode/service-override/configuration'
 import { updateUserKeybindings } from 'vscode/service-override/keybindings'
-import { createConfiguredEditor } from 'vscode/monaco'
-import { getJsonSchemas, onDidChangeJsonSchema } from 'vscode/monaco'
+import { createConfiguredEditor, getJsonSchemas, onDidChangeJsonSchema } from 'vscode/monaco'
 import { debounce } from 'throttle-debounce'
 import * as vscode from 'vscode'
 
-vscode.window.showInformationMessage('Hello', {
+await vscode.window.showInformationMessage('Hello', {
```
CGNonofr, line 20: I don't think we want to wait for it

kaisalmen: ok, sorry, that makes sense. See ⬇️

kaisalmen: README is updated, too

CGNonofr:
I'm not sure to understand how it works, why is every import of vscode/** red? 
image

also how does vite resolves the import?

kaisalmen:
It resolve the stuff from the dist folder. That's why the local code has to be built first:

```javascript
export default defineConfig({
  resolve: {
    alias: {
      'vscode/services': resolve(__dirname, './dist/services.js'),
      'vscode/service-override/messages': resolve(__dirname, './dist/messages.js'),
      'vscode/service-override/notifications': resolve(__dirname, './dist/notifications.js'),
      'vscode/service-override/dialogs': resolve(__dirname, './dist/dialogs.js'),
      'vscode/service-override/modelEditor': resolve(__dirname, './dist/modelEditor.js'),
      'vscode/service-override/configuration': resolve(__dirname, './dist/configuration.js'),
      'vscode/service-override/keybindings': resolve(__dirname, './dist/keybindings.js'),
      'vscode/service-override/textmate': resolve(__dirname, './dist/textmate.js'),
      'vscode/service-override/languageConfiguration': resolve(__dirname, './dist/languageConfiguration.js'),
      'vscode/service-override/theme': resolve(__dirname, './dist/theme.js'),
      'vscode/service-override/tokenClassification': resolve(__dirname, './dist/tokenClassification.js'),
      'vscode/service-override/snippets': resolve(__dirname, './dist/snippets.js'),
      'vscode/service-override/languages': resolve(__dirname, './dist/languages.js'),
      'vscode/monaco': resolve(__dirname, './dist/monaco.js'),
      vscode: resolve(__dirname, './dist/api.js'),
      'webpack-loader': resolve(__dirname, './dist/webpack-loader.js')
    }
  }
})
```

CGNonofr:
> ```js
> services': resolve(__dirname, './dist/ser
> ```

I see! FYI it seems you can use regexes in there or even implement a custom resolver

Can't we do something about it in the tsconfig?

kaisalmen:
> I see! FYI it seems you can use regexes in there or even implement a custom resolver

Yes, it could be written nicer. Wanted to make this work first. Question is: Is this approach better / nicer than having a separated npm package in the demo folder (as it was before)?

> Can't we do something about it in the tsconfig?

vite does not make use of a tsconfig here

CGNonofr:
> > I see! FYI it seems you can use regexes in there or even implement a custom resolver
> 
> Yes, it could be written nicer. Wanted to make this work first. Question is: Is this approach better / nicer than having a separated npm package in the demo folder (as it was before)?
> 
> > Can't we do something about it in the tsconfig?
> 
> vite does not make use of a tsconfig here

vite doesn't but vscode does for the intellisense

btw, what happens if you alias `vscode/*` to `.`? won't it use the exports in the package.json?

kaisalmen:
I updated the vite config and the tsconfig used by vscode

package.json
```diff
@@ -162,10 +167,12 @@
     "@rollup/plugin-replace": "^5.0.2",
     "@rollup/plugin-typescript": "^11.0.0",
     "@types/babel__core": "^7.20.0",
+    "@types/throttle-debounce": "~5.0.0",
     "@types/semver": "^7.3.13",
     "@types/vscode": "~1.69.1",
     "@typescript-eslint/eslint-plugin": "^5.54.0",
     "@typescript-eslint/parser": "^5.54.0",
+    "editorconfig": "~1.0.2",
```
CGNonofr, line 175: what is it for?

CGNonofr:
Does the demo code needs to be included into the main lib then? Isn't the vite aliases enough?

package.json
kaisalmen: https://editorconfig.org/

It configures tabs, line endings, see new `.editorconfig` file in root. We have that in monaco-languageclient, too.

kaisalmen:
> Does the demo code needs to be included into the main lib then? Isn't the vite aliases enough?

It is not contained in the lib. I updated the rollup config to rely on `tsconfig.build.json` which defines an exclude `src/demo`. The `tsconfig.json` is only the base config and is used by eslint/vscode.

CGNonofr:
> It is not contained in the lib. I updated the rollup config to rely on tsconfig.build.json which defines an exclude src/demo. The tsconfig.json is only the base config and is used by eslint/vscode.

But the demo dependencies still need to be included in the root dev dependencies, right?

kaisalmen:
> But the demo dependencies still need to be included in the root dev dependencies, right?

True, but they are only devDependencies.

package.json
CGNonofr: Do we need the vscode extension? I don't understand how installing a npm dependency can do anything related to the editor

CGNonofr:
> > But the demo dependencies still need to be included in the root dev dependencies, right?
> 
> True, but they are only devDependencies.

I'm not sure to see the advantage compared to having a separate folder with its own node_modules though. It's only devDependencies but it still complexify the package.json and it's not easy to know what dependency is used for what

package.json
kaisalmen: Yes, true, I forgot about that. Do want a `.github/extensions.json` or shall I revert this?

CGNonofr: What does it do? I don't see this file on monaco-languageclient

kaisalmen:
> I'm not sure to see the advantage compared to having a separate folder with its own node_modules though. 

Yes, I agree. The solution did not turn out to be simple. A npm workspace project is the better idea, but moving the lib to a sub-package only to separate it from the demo is overkill.

kaisalmen:
There are some enhancements that can be kept/incorporated like the version check and the eslint improvements

kaisalmen:
@CGNonofr I have reverted the vite or npm workspace changes and only kept enhancements (see updated list in PR description: https://github.com/CodinGame/monaco-vscode-api/pull/69#issue-1601292472)
Yesterdays status I kept on branch **kaisalmen/demo-integration-backup**

tsconfig.rollup.json
```diff
@@ -1,6 +1,11 @@
 {
-	"extends": "./tsconfig.json",
+	"extends": "./tsconfig.build.json",
+	"compilerOptions": {
+		"noEmit": false,
+		"outDir": "dist",
+		"rootDir": "./"
+	},
```
CGNonofr, line 7: I doesn't seem to change any compiler option compared to the extends

demo/package.json
```diff
@@ -4,21 +4,23 @@
...
+    "typescript": "~4.9.5",
+    "vite": "~4.1.4"
   },
   "dependencies": {
-    "monaco-editor": "~0.34.0",
-    "throttle-debounce": "^5.0.0",
-    "vscode": "npm:@codingame/monaco-vscode-api@~1.69.21",
-    "vscode-oniguruma": "~1.6.2"
+    "throttle-debounce": "~5.0.0",
+    "vscode": "npm:@codingame/monaco-vscode-api@*",
```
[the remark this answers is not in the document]
kaisalmen: Ok, then we should stick with "bugfix" level `~` as previously done:
```
 "vscode": "npm:@codingame/monaco-vscode-api@~1.69.22"
```

scripts/install-vscode
```diff
@@ -25,6 +34,9 @@ rm -rf `find . -name '*.test.ts' -o -name 'test' -o -name 'electron-browser'`
 rm -rf vs/code
 rm -rf vs/server
 
+mkdir $output_directory
```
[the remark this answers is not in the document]
kaisalmen: Goodpoint

tsconfig.rollup.json
kaisalmen: The subtle, but important inclusion change is required:
```
"./src/**/*",
```
because the main `tsconfig.json` file includes all sources (demo folder, too!):
```
"**/src/**/*",
```

kaisalmen:
```
"outDir": "dist",
"rootDir": "./",
```
can be moved to `tsconfig.json`

demo/src/main.ts
```diff
@@ -13,15 +13,16 @@ import 'monaco-editor/esm/vs/editor/standalone/browser/referenceSearch/standalon
 import 'monaco-editor/esm/vs/language/json/monaco.contribution'
 import { getUserConfiguration, onUserConfigurationChange, updateUserConfiguration } from 'vscode/service-override/configuration'
 import { updateUserKeybindings } from 'vscode/service-override/keybindings'
-import { createConfiguredEditor } from 'vscode/monaco'
-import { getJsonSchemas, onDidChangeJsonSchema } from 'vscode/monaco'
+import { createConfiguredEditor, getJsonSchemas, onDidChangeJsonSchema } from 'vscode/monaco'
 import { debounce } from 'throttle-debounce'
 import * as vscode from 'vscode'
 
+// eslint-disable-next-line @typescript-eslint/no-floating-promises
```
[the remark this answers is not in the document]
kaisalmen: Thanks, I didn't know that one

demo/package.json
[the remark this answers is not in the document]
CGNonofr: What if you use `"vscode": "file:../"`?

you'll need to change 
`import { EventEmitter } from 'vscode'` to `import { Emitter } from 'vs/base/common/event'` in `tools.ts` to make it work

tsconfig.rollup.json
CGNonofr: I'm not sure to understand, I'm talking about the compiler options only, which are the same in tsconfig.rollup.json and tsconfig.build.json

demo/package.json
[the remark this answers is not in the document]
kaisalmen: Yes, I will give that a try.

tsconfig.rollup.json
kaisalmen: I did this because:
vscode relies on a `tsconfig.json`. You just cannot configure the name of the main TypesScript config file vscode should use for whatever reasons. vscode's eslint support automatically takes into account what is defined in `tsconfig.json`.

rollup's plugin-typescript picks up automatically `tsconfig.json` (default) and therefore used demo source, too. That's why I introduced the other config.

But, I realized this is not the best way. I do it differently now. All tsconfigs are reverted. I kept the tsconfig entry in `rollup.config.ts` to make aware which tsconfig is used.
I introduced a `tsconfig.eslint.json` and added parserOptions to the eslint config.

Honestly, configuration of eslint + TypeScript is unpleasant and full of unexpected behaviour.

kaisalmen:
Pipeline fails because of this. I have no idea why suddenly 🤷‍♂️  
image

kaisalmen:
Look at this. They broke the GH action:
Last working build:
image

Broken build:
image

demo/package.json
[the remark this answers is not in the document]
kaisalmen: Done

demo/src/main.ts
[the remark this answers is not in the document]
kaisalmen: Done

package.json
kaisalmen: Discussion is outdated.

src/demo/main.ts
kaisalmen: Discussion is outdated.

package.json
kaisalmen: Discussion is outdated.

demo/package.json
kaisalmen: Discussion is outdated.

tsconfig.eslint.json
```diff
@@ -0,0 +1,17 @@
+// this tsconfig is used by @rollup/plugin-typescript
```
kaisalmen, line 1: I forgot to remove this comment

kaisalmen: FIxed ⬇️

kaisalmen:
⬆️ vite has to be run from root of project when demo uses `monaco-vscode-api` locally otherwise codicons cannot be loaded. If vite is launched from `demo` folder top level `node_modules` cannot be read. I update the start script and README, so `npm run start` does `npm i` in demo and runs vite!

Now this is good to go finally. I don't have an idea what to do about the broken GH action!?

CGNonofr:
> Pipeline fails because of this. I have no idea why suddenly man_shrugging image

Yeah I tried to install the runner on another machine but it's missing yarn...

demo/src/main.ts
```diff
@@ -13,16 +13,15 @@ import 'monaco-editor/esm/vs/editor/standalone/browser/referenceSearch/standalon
...
 import { debounce } from 'throttle-debounce'
 import * as vscode from 'vscode'
 
-vscode.window.showInformationMessage('Hello', {
+void vscode.window.showInformationMessage('Hello', {
   detail: 'Welcome to the monaco-vscode-api demo',
   modal: true
-}).then(() => {
-  vscode.window.showInformationMessage('Try to change the settings or the configuration, the changes will be applied to all 3 editors')
+}).then(async () => {
```
CGNonofr, line 23: nit: it doesn't need to be async then

package.json
```diff
@@ -179,7 +182,8 @@
     "rollup-plugin-styles": "^4.0.0",
     "semver": "^7.3.8",
     "ts-node": "^10.9.1",
-    "typescript": "^4.9.5",
+    "typescript": "^4.9.5",    
+    "vite": "~4.1.4",
```
CGNonofr, line 186: I'm not really comfortable having unrelated dependencies here, I really do think everything related to the demo shouldn't impact the main project (this file is even published)

demo/src/main.ts
kaisalmen: yes, I will update

package.json
kaisalmen: Ok, I wouldn't be this strict here, but I understand.

I added `monaco-editor` as a direct dependency to **demo** again and it works (codicons are used). There is an error in the console, but it works due to the direct dependency being used in the end. 
I hope this ok now. 🙂

.eslintrc
```diff
@@ -3,14 +3,26 @@
...
-
+  "ignorePatterns": [
+    "vscode/**/*",
+    "dist/**/*"
+  ],
+  "plugins": [
+    "@typescript-eslint",
+    "import",
+    "unused-imports"
+  ]
```
CGNonofr, line 27: I'm confused, I was thinking this project uses `@codingame/eslint-config`. It's installed but not used...

demo/package.json
```diff
@@ -4,21 +4,24 @@
...
   },
   "dependencies": {
-    "monaco-editor": "~0.34.0",
-    "throttle-debounce": "^5.0.0",
-    "vscode": "npm:@codingame/monaco-vscode-api@~1.69.21",
-    "vscode-oniguruma": "~1.6.2"
+    "monaco-editor": "~0.34.1",
+    "throttle-debounce": "~5.0.0",
+    "vscode": "file:../",
+    "vscode-oniguruma": "~1.7.0"
```
CGNonofr, line 21: I don't know if we should update it until we upgraded to monaco 0.36, I had related issues in the past

demo/src/main.ts
```diff
@@ -55,13 +54,13 @@ const settingsModel = monaco.editor.createModel(
 createConfiguredEditor(document.getElementById('settings-editor')!, {
   model: settingsModel
 })
-settingsModel.onDidChangeContent(debounce(1000, () => {
-  updateUserConfiguration(settingsModel.getValue())
+settingsModel.onDidChangeContent(debounce(1000, async () => {
+  await updateUserConfiguration(settingsModel.getValue())
 }))
-updateUserConfiguration(settingsModel.getValue())
+await updateUserConfiguration(settingsModel.getValue())
```
CGNonofr, line 60: top level awaits are probably a bad idea here

.eslintrc
kaisalmen: And I just checked, I didn't change/remove it by accident

demo/package.json
kaisalmen: Ah, yeah, this is not in line with 0.34

demo/src/main.ts
kaisalmen: Ok, then void to evade the eslint errors. Line 87 too btw

.eslintrc
CGNonofr: Nop indeed, it's probably my fault!

CGNonofr:
I tried to install the self-hosted runner on a server but I got that weird error

image

I don't have the time to investigate right now

demo/src/main.ts
kaisalmen: Done ⬇️

.eslintrc
kaisalmen: Ok, I guess, you will address this independent of this PR.

kaisalmen:
FYI, I saw this this morning, too, but it worked then after a runner picked it:
image

kaisalmen:
I have tested the build on a fork I created this morning and used ubuntu-latest as runner and it works nicely:
https://github.com/kaisalmen/monaco-vscode-api/actions/runs/4311997614/jobs/7522012650

CGNonofr:
> I have tested the build on a fork I created this morning and used ubuntu-latest as runner and it works nicely: https://github.com/kaisalmen/monaco-vscode-api/actions/runs/4311997614/jobs/7522012650

You mean you use the official runners? I switched to hosted one because there wasn't enough memory and it was crashing.

kaisalmen:
> You mean you use the official runners?

Yes, we don't have self-hosted runners and I forked it under my GH account

kaisalmen:
Did you (I mean someone at your company) update the runners recently, like yesterday 😉 ?

CGNonofr:
> Did you (I mean someone at your company) update the runners recently, like yesterday wink ?

What do you mean? I have to run the runner on my own machine. yesterday I tried to create a docker of it to make it run on one of our dev server but I failed for a mysterious reason.

Can you try going back to official runner here?

kaisalmen:
> What do you mean? I have to run the runner on my own machine. yesterday I tried to create a docker of it to make it run on one of our dev server but I failed for a mysterious reason.

Oh, really on your own machine!? Any idea, when you have time to fix it. Jobs aren't currently picked up. The job from yesterday  evening was not picked up over night and I tried to re-run with no luck, so far.

> Can you try going back to official runner here?

I did it on a fork for testing only, nothing changed here and I do not intend to 🙂

CGNonofr:
> Oh, really on your own machine!?

Yeah, that's the only project that requires it and I didn't take the time to run it elsewhere, is required a private token so I can't just run it on any machine, it needs a ansible conf, I tried yesterday without success.

> I did it on a fork for testing only, nothing changed here and I do not intend to

feel free to do it if it works now!

kaisalmen:
> feel free to do it if it works now!

Ok, I will try it now

CGNonofr:
> It already worked without doing anything

Yes I restarted the runner on my machine

kaisalmen:
This PR has became way to lengthy. 😆 

Are you ok with the changes now? Runner config adjustments (if you want any) should be done independently, I think.

CGNonofr (approved):

PR #73, opened 2023-03-02
Use official runners

CGNonofr:


CGNonofr:
@kaisalmen you were right, it works on official github runners...

kaisalmen (approved):
LGTM. Why are these eslint settings no longer required?

CGNonofr:
> LGTM. Why are these eslint settings no longer required?

There is an extends and those configs are already in the extended config

kaisalmen:
> There is an extends and those configs are already in the extended config

Yes, thanks, just checked in parallel. That is all defined in `@codingame/eslint-config`

PR #81, opened 2023-03-21
Fail pipeline when the publish fails

CGNonofr:


CGNonofr:
@kaisalmen I see no reason for the publish to fail, it wasn't changed in the previous PR. This change will allow us to retry the publishing

kaisalmen (approved):
LGTM

kaisalmen:
Ok, let's observe if it happens again.

PR #84, opened 2023-03-22
Debug services

CGNonofr:
This PR includes 2 main changes:

## It changes the way vscode extension are loaded [BREAKING CHANGE]

Instead of providing languages, grammars, themes etc... to each service-override modules, now you can call `registerExtension` from `vscode/extensions` which takes as parameter the package.json (containing the contribute section) of the extension and return 2 things:
- the vscode api designed for that extension (currently we were only able to get the extension for a single default extension)
- a `registerFile` method to set the content of the file references in the package.json of the extensions (themes, grammars...)

I think it make the interface clearer (the user doesn't have to think about schemes of its extensions files, everything is in the `extension` scheme and prefixed by an UUID, the user only manipulate paths

It's also a step toward a feature that would allow to load a vscode extension as is

## It introduces the vscode debug services

It allows to use the [debug adapter protocol](https://microsoft.github.io/debug-adapter-protocol/) on monaco

It is included it in the demo using `graalvm` and `docker` which implements the protocol for its node runtime, but as it requires a server, it won't work on https://codingame.github.io/monaco-vscode-api/


kaisalmen:
@CGNonofr wow, cool. That's what you meant. I will try to test it tomorrow.

CGNonofr:
@kaisalmen I've release a [1.76.3-next.0](https://www.npmjs.com/package/@codingame/monaco-vscode-api/v/1.76.3-next.0)

Will you be able to have a look at it?

kaisalmen:
and hopefully overcome the build issues (webpack verification/ angular example) in monaco-languageclient as well, because I will use this version for testing

demo/src/setup.ts
```diff
@@ -42,13 +46,84 @@ window.MonacoEnvironment = {
...
+}
+
 // Override services
 StandaloneServices.initialize({
-  ...getModelEditorServiceOverride(async (model, options) => {
-    // eslint-disable-next-line no-console
-    console.log('trying to open a model', model, options)
-    return undefined
-  }),
+  ...getModelEditorServiceOverride(openNewCodeEditor),
```
kaisalmen, line 126: Can you explain what the purpose of this `openNewCodeEditor` is`?

kaisalmen (commented):
@CGNonofr I was able to build everything locally. The demo works fine. I was able to run the debugServer and execute the debugger (saw the console feedback). Only thing I needed to do beforehand was to manually pull the docker image.

A README explaining what the demo contains and what it does could be helpful.

Also, regarding my question. It wo

kaisalmen:
@CGNonofr I haven't yet looked more closely at the code, but what I took from it already is that you want (from this PR forward) to fully rely on the vscode extension registration mechanism, so people already familiar with using the vscode extension API will find adaptation easy. Correct?

Regarding `monaco-languageclient` it makes perfect sense to base v5 on this. Although the library is not directly impacted.

demo/src/main.ts
```diff
@@ -88,7 +91,8 @@ const settingsModel = monaco.editor.createModel(
   "editor.semanticHighlighting.enabled": true,
   "editor.bracketPairColorization.enabled": false,
   "editor.fontSize": 12,
-  "audioCues.lineHasError": "on"
+  "audioCues.lineHasError": "on",
+  "audioCues.onDebugBreak": "on"
```
kaisalmen, line 95: Should setting break points work? I know this about audio config, but it implies break points should work. It does not here.

kaisalmen (commented):
The list of changes is too long for a detailed review in reasonable amount of time. I have scanned through it just now.
But, I think that the general approach you have taken is very good. The PR is actually far more than only the addition of debug services. 🙂 

I really recommend to update the main README and to add an explanation to the demo either with this PR or with another one.
Should setting break points already and halting the execution work?

If you want to add nothing now, I will approve the changes soon. Otherwise I review additional additions as soon as they are ready.

demo/src/setup.ts
CGNonofr: It's a function that is called when a new editor should be opened to display a model given as parameter

It happens for instance when you ctrl+click on an import of another file with LSP enabled (and when there is a file content provider registered)

It can also happen with the debugger: try to enter the `console.log` call while debugging and you'll see another editor in a popup containing a node source code file

demo/src/main.ts
CGNonofr: What do you mean? You should be able to add breakpoint in the first editor, maybe I broke it?

CGNonofr:
> @CGNonofr I was able to build everything locally. The demo works fine. I was able to run the debugServer and execute the debugger (saw the console feedback). Only thing I needed to do beforehand was to manually pull the docker image.

Nice! Indeed the script should probably pull the image at startup, forgot it!

> A README explaining what the demo contains and what it does could be helpful (see the question regarding `openNewCodeEditor`).

Yeah, documentation is not my passion :) I'll add some!

> I haven't yet looked more closely at the code, but what I took from it already is that you want (from this PR forward) to fully rely on the vscode extension registration mechanism, so people already familiar with using the vscode extension API will find adaptation easy. Correct?

Exactly! it's really more straighforward to use a vscode extension "as is" now. Note that this changed was required for the debugger because some vscode code was requesting the extension manifest directly instead of relying on the extension point.

I also have some plan to have a function that takes a vsix directly and load it, but I don't know exactly how to do it without requiring to have the whole extension (containing binaries and unused javascript code) on the client

> This is awesome work. IMO.

Thank you! I'm quite happy with the result

> The list of changes is too long for a detailed review in reasonable amount of time. I have scanned through it just now.
But, I think that the general approach you have taken is very good. The PR is actually far more than only the addition of debug services. 🙂

Indeed! it was required for it. It would have been a good idea to do it in 2 PRs but I realized I needed it in the middle and now it's pretty hard to split it

> I really recommend to update the main README and to add an explanation to the demo either with this PR or with another one.

Will do! feel free to do it if you find the inspiration :)

> Should setting break points already and halting the execution work?

Yeah! that's the main feature

CGNonofr:
@kaisalmen 

https://user-images.githubusercontent.com/23257479/228386708-5bae4a7c-41d7-476c-a2c8-bf571d7dedaf.mp4

CGNonofr:
@kaisalmen added 2 fixes:
- df9a5c1344e1bd5c98a4df58f658e6a589b3a2d4 to pull the docker image
- 58f8f515362ac19842d3e149357ab9ebe92164fb to keep the `memoize` decorator on class methods (it has an impact on breakpoints, maybe that's what you were encountering)

kaisalmen:
@CGNonofr debugger works. I use a Ubuntu WSL2 for this repo and I need to use the Linux browser, otherwise there seems to be some sort of network resolution problem when I use a Windows browser.

kaisalmen (approved):
Feel free to merge. Let's defer the README adjustments,

kaisalmen:
> Yeah, documentation is not my passion :) I'll add some!

I know that one. My time for open-source work is a bit limited this week. I hope to do more next week (documentation + monaco-languageclient)

PR #86, opened 2023-04-03
Fixes webpack 5 esm build problems

kaisalmen:
This fixes a problem with webpack 5.
In monaco-languageclient the webpack 5 build verification is broken. It builds esm code and does not like these two imports.


CGNonofr (approved):
Thanks!

PR #94, opened 2023-04-07
Update to monaco 0.37

CGNonofr:
based on https://github.com/CodinGame/monaco-vscode-api/pull/93

PR #96, opened 2023-04-11
Expose default vscode extensions

CGNonofr:
based on https://github.com/CodinGame/monaco-vscode-api/pull/94

the package now exposes some module that register vscode default extensions:

for instance:
```typescript
import 'vscode/default-extensions/theme-defaults'
import 'vscode/default-extensions/javascript'
```

is now enough to have the default vscode themes and the javascript language


kaisalmen:
This sounds good. I will review it tomorrow.

CGNonofr:
@kaisalmen I've also added a rollup/vite plugin that allows to load a vsix file directly (only the manifest or course, not the code)

kaisalmen:
@CGNonofr I thought this is independent, but now I looked at the git graph and I have to start with
https://github.com/CodinGame/monaco-vscode-api/pull/93 right?

CGNonofr:
> @CGNonofr I thought this is independent, but now I looked at the git graph and I have to start with #93 right?

Yep, every PR is based on the previous one

Alsmost everytime I make changes I need to tweak the polyfills so I prefer to be based on the last working version to not have to rebase all remaining PRs and test them when one is merged

CGNonofr:
I've published a [1.78.0-next.0](https://www.npmjs.com/package/@codingame/monaco-vscode-api/v/1.78.0-next.0) including all the open PRs

kaisalmen:
@CGNonofr Doesn't it make sense to only look at the delta between main and #98 because otherwise I check things that are gone a couple of commits later. WDYT?

CGNonofr:
> eworked a couple of comm

The PRs are based on each other but quite still independent. I don't think there is anything done then undone, the commits are mostly well splitted. I'm not sure you'll be able to understand everything by just looking at the whole changes.

But It's a matter of preferences, do as you wish

README.md
```diff
@@ -135,15 +140,62 @@ npm install -D @types/vscode
...
+```
+
+### Loading vsix file
+
+VSCode extension are bundled as vsix files.
+This library exposes a rollup plugin (vite-compatible) that allows to load a vsix file. The code is not used, only the declarative part in the manifest.
+
+- rollup/vite config:
+```typescript
+import vsixPlugin from 'vscode/rollup-vsix-plugin'
```
kaisalmen, line 187: Have you already made use of it somewhere?

kaisalmen (approved):
I looked at the real delta locally and let's go. 🎉

kaisalmen:
> The PRs are based on each other but quite still independent. I don't think there is anything done then undone, the commits are mostly well splitted. I'm not sure you'll be able to understand everything by just looking at the whole changes.

Yes, looking at them separately, but with an external tool to see the real delta in between the branches, was the better approach.

README.md
CGNonofr: I did on the demo with the ocaml vscode extension. The demo already contains a lot of stuff and I don't really know how to showcase it properly

PR #103, opened 2023-04-20
Fix workspace warning log

CGNonofr:


kaisalmen (approved):
In this case throwing an error here makes no difference to `Promise.reject`

Issue #113, opened 2023-05-12
Treemending failing


MrMarble:
Trying to setup monaco-vscode-api with pnpm fails to "treemend"

Here are the logs of a clean (removed node_modules folder) install:

```
$pnpm i

Lockfile is up to date, resolution step is skipped
Packages: +1246
++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Packages are hard linked from the content-addressable store to the virtual store.
  Content-addressable store is at: F:\.pnpm-store\v3
  Virtual store is at:             node_modules/.pnpm
Downloading registry.npmjs.org/monaco-editor/0.37.1: 16 MB/16 MB, done
Progress: resolved 1246, reused 1243, downloaded 2, added 1246, done

dependencies:
+ @monaco-editor/react 4.5.1
+ @tauri-apps/api 1.3.0
+ monaco-editor 0.37.1
+ react 18.2.0
+ react-dom 18.2.0
+ react-resizable-panels 0.0.42
+ react-use 17.4.0
+ tauri-plugin-store-api 0.0.0
+ uuid 9.0.0
+ vscode <- @codingame/monaco-vscode-api 1.78.5
+ zustand 4.3.8

devDependencies:
+ @storybook/addon-essentials 7.0.10
+ @storybook/addon-interactions 7.0.10
+ @storybook/addon-links 7.0.10
+ @storybook/addon-styling 1.0.6
+ @storybook/blocks 7.0.10
+ @storybook/react 7.0.10
+ @storybook/react-vite 7.0.10
+ @storybook/testing-library 0.0.14-next.2
+ @types/mocha 10.0.1
+ @types/node 20.1.0
+ @types/react 18.2.6
+ @types/react-dom 18.2.4
+ @types/uuid 9.0.1
+ @types/vscode 1.78.0
+ @typescript-eslint/eslint-plugin 5.59.2
+ @typescript-eslint/parser 5.59.2
+ @vitejs/plugin-react 4.0.0
+ @wdio/cli 8.10.0
+ @wdio/local-runner 8.10.0
+ @wdio/mocha-framework 8.10.0
+ @wdio/spec-reporter 8.10.0
+ autoprefixer 10.4.14
+ eslint 8.40.0
+ eslint-config-prettier 8.8.0
+ eslint-plugin-react-hooks 4.6.0
[...]

PR #115, opened 2023-05-15
Don't apply treemending patch if file hashes match

kaisalmen:
After treemending patch creation a file including names and hashes of all files of the esm folder is created (expectation).
This expectation file is added to the `monaco-vscode-api` package.
When `monaco-treemending` is executed it performs the same calculation on the installed monaco-editor esm files and successfully terminates early if the file hashes meet the expectation.

This should fix https://github.com/TypeFox/monaco-languageclient/issues/491 and #113 , but a test with pnpm is still outstanding

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,5 +2,4 @@ node_modules/
 vscode/
 vscode-default-extensions/
 dist/
-monaco-editor-treemending.patch
-monaco-editor-hashes.txt
+monaco-editor-treemending.*
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ vscode/
 vscode-default-extensions/
 dist/
 monaco-editor-treemending.patch
+monaco-editor-hashes.txt
--- a/package.json
+++ b/package.json
@@ -157,8 +157,7 @@
   "files": [
     "dist",
     "monaco-treemending.js",
-    "monaco-editor-treemending.patch",
-    "monaco-editor-hashes.txt",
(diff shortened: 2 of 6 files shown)


CGNonofr:
Do we need to compute the hash of every single file? I mean, there is 3 cases:
- The monaco-editor version is wrong
- The monaco-editor version is correct and the patch **wasn't** already applied
- The monaco-editor version is correct and the patch **was** already applied

Can't we just check the version in the package.json for the first case, and touch a file or modify the package.json of monaco-editor to store that we've patched it?

CGNonofr:
What do you think about https://github.com/CodinGame/monaco-vscode-api/pull/116 ?

kaisalmen:
@CGNonofr  Relying on a third party tool may be a good idea: https://github.com/TypeFox/monaco-languageclient/issues/492#issuecomment-1562433613

CGNonofr:
> @CGNonofr Relying on a third party tool may be a good idea: [TypeFox/monaco-languageclient#492 (comment)](https://github.com/TypeFox/monaco-languageclient/issues/492#issuecomment-1562433613)

Please tell me how then, I've tried a lot of things already before implementing the current manual way

kaisalmen:
> Please tell me how then, I've tried a lot of things already before implementing the current manual way

I thought it would help in applying an preventing errors, but exactly here it is not really helping. I just played around with it. 👎

kaisalmen:
We will go with #116

PR #119, opened 2023-05-31
Introduce treemending work-around for pnpm

kaisalmen:
@CGNonofr Patching directly doesn't work with pnpm. Every file of every used node_modules is contained in a data store. Patching files directly does not work. I debugged it locally, during the patch process suddenly files have already been patched and the patch process stops. My wild guess is that the file traversal is not consistent. 🤷‍♂️ 

Anyway, when a `.pnpm` directory is detected in monaco-editor path this workaround copies all esm files to a tmp dir patches everything and copies files back. This works in nicely in my test pnpm project.

### DIFF
--- a/src/monaco-treemending.ts
+++ b/src/monaco-treemending.ts
@@ -1,6 +1,7 @@
 import { applyPatches, ParsedDiff } from 'diff'
 import * as fs from 'fs/promises'
 import * as path from 'path'
+import * as os from 'os'
 import { createRequire } from 'node:module'
 const require = createRequire(import.meta.url)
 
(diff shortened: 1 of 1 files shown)


src/monaco-treemending.ts
```diff
@@ -67,6 +77,11 @@ async function run () {
     })
   })
 
+  // in case of pnpm copy the temp esm back
+  if (monacoEsmDirectory.includes('.pnpm')) {
```
CGNonofr, line 81: nit: do we need to check if the file exist twice?

CGNonofr:
I'm not familiar with pnpm, how does it help patching it elsewhere?

kaisalmen:
It uses a datastore in your home directory, so you don't have node_modules twice. The files in node_modules are linked/virtual. This makes problems during the patch process. Copying forth and back seems reliable. I came to this approach because `pnpm patch` creates a tmp directory where you apply changes and then the commit step moves it back.

It don't really like tool specific work arounds, but this is the one with least impact, IMO

kaisalmen:
Btw, with this in place treemending as postinstall works with monaco-languageclient with npm, yarn and pnpm.

CGNonofr (dismissed):
Feel like a hack to me, I hope it won't introduces other issues

maybe the patch approach is still not the best way...

Let's merge it and think about a better solution

src/monaco-treemending.ts
kaisalmen: No, you are right. Give me a minute

kaisalmen:
Resolution is not the problem, but file traversal. I always ended up with files that were already patched although they were untouched before. It is really weird. It patched let's say 50 files and suddenly one was already patched. Ahhh, maybe it is contained twice and then the datastore stores it only once and then boom

CGNonofr:
reading about [package-import-method](https://pnpm.io/npmrc#package-import-method), it seems that by default it used a cloning mode, is this the issue? doesn't it fail when it's not supported? it says it doesn't work on mac

kaisalmen:
@CGNonofr It works fine with https://www.npmjs.com/package/monaco-languageclient/v/6.0.3-next.4

PR #120, opened 2023-06-02
Editor and status bar components

CGNonofr:


kaisalmen:
@CGNonofr Please rebase and signal when it is ready for review / test. Thanks

CGNonofr:
> @CGNonofr Please rebase and signal when it is ready for review / test. Thanks

Rebased and demo build fixed :)

src/monaco.ts
```diff
@@ -142,14 +154,69 @@ export async function createModelReference (resource: URI, content?: string): Pr
   return (await StandaloneServices.get(ITextModelService).createModelReference(resource)) as IReference<ITextFileEditorModel>
 }
```
kaisalmen, line 157: For my understanding: Can you explain why you did this: "refactor: only use StandaloneKeybindingService for editors"
Would this lead to too many unused keyboard services?

kaisalmen: One cannot write a review comment against a specific commit message as far as I know, that's why I attached it to some code.

kaisalmen (approved):
Nice work. It works as expected. I think it make sense to include this along with monaco-editor 0.38.0 into the upcoming monaco-languageclient release 6.1.0

src/monaco.ts
CGNonofr: There will be only 2 services, the Standalone and the Workbench, just like before.

The only difference is that the standalone one will only be used by Standalone editors (required so the `addAction` api still works) and the Workbench for everything else. The workbench quickaccess code requires to have access to the Workbench keydinging service directly

kaisalmen: Thank you.

Issue #122, opened 2023-06-04
Wasm import causes library to break with Vite


TaeHagen:
When using this library with Vite it crashes stating `[vite] Internal server error: "ESM integration proposal for Wasm" is not supported currently. Use vite-plugin-wasm or other community plugins to handle this.`

The problematic line is: https://github.com/CodinGame/monaco-vscode-api/blob/4ab2407733597b8a543331c4f6687650b57ce241/src/service-override/textmate.ts#L11

Vite interprets this as a WASM ESM module and throws an error. Attempting to install `vite-plugin-wasm` results in errors about unresolved imports as the WASM file is not an ESM module. Appending a `?url` to the link causes the import to instead resolve to the URL when using Vite ([docs](https://vitejs.dev/guide/features.html#static-assets)) and the library works as intended.

I'm not sure how you would want to handle Vite, but as far as I can tell there isn't any other way to fix this other than appending a `?url`. At this point I have to add it manually in my node_modules which is really annoying.

This is a really useful library and I'm looking forward to using it in my Vite projects!

CGNonofr:
Fyi, the demo here uses vite as well.

You need to add `assetsInclude: ['**/*.wasm']` un you config

TaeHagen:
Thanks! I guess I need to look at the vite documentation again :)

CGNonofr:
You're right, it's not mentionned! :) feel free to contribute to the documentation ;)

PR #123, opened 2023-06-04
Update installation instructions for Vite

TaeHagen:
For https://github.com/CodinGame/monaco-vscode-api/issues/122

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,11 @@ npm install -D @types/vscode
 }
 ```
 
+If you use Vite, add in your vite.config.js:
+```js
+assetsInclude: ['node_modules/vscode-oniguruma/**/*.wasm']
+```
+
 ### Why?
 
 Monaco-editor is a library that is constructed using code from vscode and goes through an intense treeshaking process.


README.md
```diff
@@ -26,6 +26,11 @@ npm install -D @types/vscode
 }
 ```
 
+If you use Vite, add in your vite.config.js:
+```js
+assetsInclude: ['**/*.wasm']
```
CGNonofr, line 31: It works well in the demo here because there is nothing else, but it may break application already using wasm via the wasm loader. Maybe the glob should be more specific, don't you think?

CGNonofr (changes requested):

CGNonofr:
Thank you for your contribution! :)

README.md
TaeHagen: I changed it to `node_modules/vscode-oniguruma/release/onig.wasm`. This works right now but could break with library updates if the path were to change.

CGNonofr: Maybe you can includes all wasm inside vscode-oniguruma then?

TaeHagen: I don't know why I didn't think of that. Great idea!

CGNonofr:
Can you please follow the semantic release convention for commit messages?

TaeHagen:
Done :) I squashed the commits and updated the message.

CGNonofr (approved):

PR #130, opened 2023-06-19
Do not provide blob urls to json worker

CGNonofr:
It fails when there is multiple schemas for the same association, the worker merges them by creating a  `allOf` parent schema that fails with blob url because it tries to resolve them as paths

I got headaches on this one


kaisalmen (approved):
Sounds logical, but I bet it was hard to find the root cause. How did you stumble upon the problem in the first place?

CGNonofr:
> How did you stumble upon the problem in the first place?

I've been using the vsix plugin to load some vscode extensions and multiple ones where registering json schemas for the package.json file. And as soon as there is more than 1, the json worker create a combined one and that's where it starts to fails with resolution failure and attempt to use fetch to get the underlying schema on a fake url with scheme `schemaservice://`

Issue #136, opened 2023-06-28
Bundle size when using webpack


char0n:
Hi again,

When updating from `v1.76.7` to `v1.79.6` I've noticed that the JavaScript bundle size of this library when built with webpack@5 bumped from `1.05 MB` to `1.55 MB`. It's half of megabyte of additional code. I've started to hit the JavaScript memory limits. I know it can be extended, but I'm looking for a way how to use this library only with minimal code required to integrate with `monaco-editor`.

Here are all the imports I currently use:

```js
import { initialize as initializeExtensions } from 'vscode/extensions';
import { StandaloneServices, IStorageService } from 'vscode/services';
import { languages as vscodeLanguages } from 'vscode';
```

On `vscodeLanguages` I call the following methods:

- createDiagnosticCollection
- register*Provider
- setLanguageConfiguration

---

I've tried to `hack` this library and set `"sideEffects": false` in `package.json` to allow webpack@5 to tree-shake as much as possible. I've created the new `api1.js` file and mapped it via `package.json` exports fields:

**api1.js**
```
import create$5 from './vscode-services/languages.js';
import { getDefaultExtension } from './default-extension.js';

export const languages = create$5(getDefaultExtension)
```

Then I use it in following way:

```js
import { languages as vscodeLanguages } from 'vscode/api1';
```

Of course it didn't help and I'm getting:

![image](https://github.com/CodinGame/monaco-vscode-api/assets/193286/a57ac762-c51c-4fb6-a9f4-fe94dea9b6cc)

Any advice please?

---

**Memory limits hit:**

[2735:0x73a8570]   271065 ms: Mark-Compact 2041.4 (2087.3) -> 2038.2 (2088.1) MB, 2524.44 / 0.00 ms  (average mu = 0.079, current mu = 0.011) allocation failure; scavenge might not succeed
[2735:0x73a8570]   274485 ms: Mark-Compact 2042.2 (2088.1) -> 2039.1 (2097.1) MB, 3408.39 / 0.00 ms  (average mu = 0.037, current mu = 0.003) allocation failure; scavenge might not succeed

<--- JS stacktrace --->

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: 0xc925e0 node::Abort() [/opt/hostedtoolcache/node/20.3.1/x64/bin/node]
 2: 0xb6d969  [/opt/hostedtoolcache/node/20.3.1/x64/bin/node]
[...]

PR #137, opened 2023-07-01
Optional ExtHosts - improve codesplitting/treeshaking

CGNonofr:
based on https://github.com/CodinGame/monaco-vscode-api/pull/134

related to #136

It allows to only import some parts of a api.

It would allow for monaco-languageclient to only import what is required from the api by importing `vscode/ext-hosts/languageClient` only

(that's only 14 parts out of 26)


src/override/vs/workbench/browser/parts/editor/textEditor.weak.js
```diff
@@ -0,0 +1,17 @@
...
+  static [Symbol.hasInstance](instance) {
+      return AbstractTextEditor != null && instance instanceof AbstractTextEditor
+  }
+}
+
+export function setAbstractTextEditor(_AbstractTextEditor) {
+  AbstractTextEditor = _AbstractTextEditor
+}
+
+export { FakeAbstractTextEditor as AbstractTextEditor }
```
CGNonofr, line 17: @kaisalmen I didn't find any better way to do it, any other/more automatic idea?

kaisalmen: @CGNonofr I Don't have a better idea

kaisalmen (dismissed):
LGTM. I need to test the full stack with `monaco-languageclient`, so the approval of #146 is the one that counts.

CGNonofr:
> LGTM. I need to test the full stack with `monaco-languageclient`, so the approval of #146 is the one that counts.

Yeah sure, we'll only merge the last one when everything will be reviewed

PR #139, opened 2023-07-04
expose output service and more editor primitives

### LINKED ISSUE #138: Exposing VSCode Types
First I'd like to say that this project is incredibly impressive. I'm amazed that you've been able to create an _extensible_ package from the VSCode codebase and expose it as a library.

I'm currently exploring this library, and I'm wondering about the service override story. I'm trying to override a service with a custom implementation, but in some cases I need to use VSCode types that are currently not exposed by `monaco-vscode-api`. Specifically, I'm now looking how I can override `IOutputService`, but the interface is not available from the types alone.

So I'm wondering, would it be possible to expose the types from VSCode in the package as well? That would make it easier to override services with custom ones, without having to pull in the VSCode codebase.

CompuIves:
There's also a `.codesandbox/Dockerfile` in there, which I used to make it work with CodeSandbox. I'm also happy to remove it if you prefer!

This PR exposes `IOutputService` and many more interfaces from the editor, in case you want to create your own `IEditorGroupsService`. I was contemplating whether I should just do `export * from '...'`, but went with the safe route for now :).

Fixes #138

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,3 @@
+FROM node:18-bullseye
+
+RUN apt update -y && apt install -y jq rsync libsecret-1-dev
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,7 @@
+{
+  "name": "monaco-vscode-api",
+  "build": {
+    "context": ".",
+    "dockerfile": "./Dockerfile"
+  }
+}
--- a/src/services.ts
+++ b/src/services.ts
@@ -88,7 +88,48 @@ export { IProductService } from 'vs/platform/product/common/productService'
 export { IQuickInputService } from 'vs/platform/quickinput/common/quickInput'
 export { ITelemetryService } from 'vs/platform/telemetry/common/telemetry'
 export { IUriIdentityService } from 'vs/platform/uriIdentity/common/uriIdentity'
-export { IEditorGroupsService } from 'vs/workbench/services/editor/common/editorGroupsService'
+export {
+  IEditorGroupsService,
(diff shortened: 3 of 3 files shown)


CGNonofr:
Thank you!

Can you please follow the semantic release convention for your commits? (and splitting it in 2 commits - types and dockerfile) to make the automatic changelog relevant

How does it work for codesandbox though? I'm not familiar with it

CompuIves:
Okay! It's now split up in two commits.

> How does it work for codesandbox though? I'm not familiar with it

I've now added a [devcontainer](containers.dev) configuration for this repo. This essentially tells what environment this repo would need to properly run, which is useful for local development (VSCode can hint suggestions because of it), but it also helps for online development.

CodeSandbox is one of those online editors (I'm one of the co-founders, so there's bias). It allows you to open the repo in an online code editor, which already runs the repo so you don't have to clone locally and set up the environment yourself. Here's an example: https://codesandbox.io/p/github/CodinGame/monaco-vscode-api/main, although you can see it has an error currently because it lacks the configuration this PR contributes 😄 .

There are two advantages to it:

1. Every PR has a running dev env that you can use for reviewing, so you don't have to check it out locally
2. Contributors can open the online editor, click "Create Branch" and contribute more easily back to the repo

I've changed the CodeSandbox configuration to become a devcontainer configuration, so it will work with all online editors that use it (Codespaces as well, for example).

CGNonofr:
Ok, looks great! thanks!

CGNonofr (approved):

CGNonofr:
https://codesandbox.io/p/github/CodinGame/monaco-vscode-api/main seems to still be not working, jq is required :)

PR #141, opened 2023-07-07
feat: expose SimpleEditorService

CompuIves:
Here I am again! Getting closer to implementing our own tab system, but we decided that we will use Monaco editors first, inside the tabs. For this, we need to return an EditorPane, and I saw that you already built this nice wrapper which exposes `EditorPane`. I initially tried copying it to our codebase, but there are some functions inside VSCode that we cannot import (`isPreferredGroup`, `SIDE_GROUP`). I thought this would be easier for us, but let me know if you prefer it to be handled differently.

### DIFF
--- a/src/service-override/editor.ts
+++ b/src/service-override/editor.ts
@@ -3,16 +3,18 @@ import { IEditorOverrideServices, StandaloneServices } from 'vs/editor/standalon
 import { IResolvedTextEditorModel, ITextModelService } from 'vs/editor/common/services/resolverService'
 import { ICodeEditorService } from 'vs/editor/browser/services/codeEditorService'
 import { CodeEditorService } from 'vs/workbench/services/editor/browser/codeEditorService'
-import { IEditorService, PreferredGroup } from 'vs/workbench/services/editor/common/editorService'
-import { IEditorPane, IResourceDiffEditorInput, ITextDiffEditorPane, IUntitledTextResourceEditorInput, IUntypedEditorInput } from 'vs/workbench/common/editor'
+import { IEditorService, IRevertAllEditorsOptions, ISaveAllEditorsOptions, ISaveEditorsOptions, ISaveEditorsResult, PreferredGroup, SIDE_GROUP_TYPE } from 'vs/workbench/services/editor/common/editorService'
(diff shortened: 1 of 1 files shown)


CGNonofr:
What about exposing the `EditorService` instead, so you can override it?

Btw, what feature is missing from it that requires you to implement your custom implementation?

CompuIves:
Good one, that works too. I could even use the `openEditor` of that service.

> Btw, what feature is missing from it that requires you to implement your custom implementation?

No specific feature missing, what I'm trying to do is quite custom. This is our editor:
image

As you can see we already have a tab system (with tab splitting, etc...). I'm looking if we can implement a custom EditorGroupsService that will work with _our_ tab system. So that other services like `IHistoryService` can interact with our tab system. This remains an experiment (how easy is it to do?). If it doesn't work, I'm considering replacing our editor pane with the VSCode editor part.

src/service-override/editor.ts
```diff
@@ -98,5 +98,6 @@ export {
   OpenEditor,
   IEditorOptions,
   IResolvedTextEditorModel,
-  IReference
+  IReference,
+  EditorService
```
CGNonofr, line 102: If it's exposed, we should probably give it a less generic name, like `SimpleEditorService` or something, don't you think?

CompuIves: Good one, will do!

CGNonofr: Just change the commit message and we're good to go :)

CGNonofr:
Oh there is some warnings, as the class is exported, it needs the return types and parameters to be explicit

also if you want to be able to override methods, you need the class to repeat the parameters (prefixed by a `_` to not have a warning because unused)

CompuIves:
Fixed the TS case, but couldn't get the params to work (it still complains):
image

For me, I mainly need access to `wrapOpenEditor`, as I need to call it from `openEditor` in an `EditorGroup`. An alternative is that I use `editorService` directly and call `openEditor` on that one, but for that to work I will need to add two more events to `editorService` (`onDidEditorOpenFail` and another one), because the `IHistoryService` expects that for some reason.

So for me merging this will be already enough, or exposing `wrapOpenEditor` separately or adding the two events to `editorService`. Do you have a preference?

CGNonofr:
The issue was due to the eslint rule not being on main yet but on an open PR

I've put the fix in [another PR](https://github.com/CodinGame/monaco-vscode-api/pull/142)

CGNonofr:
Are you ok with closing that PR? Are the changes fine?

CompuIves:
Ahh, thanks a bunch for doing the work on this! Absolutely fine with closing the PR.

PR #146, opened 2023-07-12
Use extension service

CGNonofr:
based on https://github.com/CodinGame/monaco-vscode-api/pull/143

bug: allow to use the extension api even if there is no `browser` field set


CompuIves:
Very interesting! Does this also mean that the extensions run in a worker. We've been talking about using nodebox (https://nodebox.codesandbox.io/) to run VSCode extensions in the browser.

CGNonofr:
> un VSCode extensions in the brows

Not yet, but it's a big step toward making it possible by properly splitting extHost and mainThread components

There is 2 mains issues yet to be solved:
- vscode use an iframe that starts a worker for some reason, I need to find a way to make the iframe works as it try to loads an urls built on the fly... or either get rid of the iframe
- we still need to have the api object in the browser (not in the worker)

Why is it important for you that extensions run in a worker?

CompuIves:
> Why is it important for you that extensions run in a worker?

Main reason is performance, making sure that the extensions don't hold up the main thread while loading. The other one is security (in the future), if we're going to run untrusted extensions then they should not have access to the main window. This might also be why VSCode runs them in an iframe, actually.

Nodebox also runs in an iframe by default, so that might be a solution. It could perhaps get a bundle of the extension host, and execute it. It supports things like `child_process.spawn` etc as well 🤔 .

In our "old" editor (codesandbox.io/s/new), we run the extension host in a web worker. This is a very old version of VSCode though (from 2019), not nearly as polished as this package.

> we still need to have the api object in the browser (not in the worker)

Yes, that's also interesting. VSCode _does_ support having two extension hosts (browser + node), right? I wonder if we can add arbitrary amount of extension hosts...

digitallyserviced:
@CGNonofr i cannot believe I stumbled upon this!

I have been **_trying_** to port few contrib/feats from vscode to monaco and always end up running into shit i have to rip out like dependencies / injections / services / media / theming, to get it half working. 

Even then I end up having to hobble the feature im trying to actually use... because of the absolutely piss poor code/file structure of vscode that I simply dont have the time or care for to work magicly like you did!

My plans for the editor were small to begin with, but then i wanted to add things, and found that vscode's source is just WTF with junk docs, and no real care for monaco from anybody with decent examples, documentation, or creative usage other than a few projects on GH... Theia is better than vscode i feel lol 

Im so glad you've reworked some of these issues into a less coupled easily overridable way that I'm not tracking shit over 20 directories just to get the gist of what I want to do being overly time consuming to implement.

Love the demo's setup! Dunno if I can sponsor but will try... Great work on this branch

kaisalmen:
@CGNonofr you have to rebase the whole commit stack, maybe after #150 is ready?

kaisalmen:
@CGNonofr look what the monaco-languageclient webpack verification build revealed:
image

kaisalmen:
> Yep I'm aware of this issue and it's fixed in the upcoming PR, do you want me to backport it?

If the upcoming PR is based on this, then please release a new next version, then there is no need for backporting as I can directly test the new next version with monaco-languageclient.

PR #147, opened 2023-07-13
fix: expose bulkEditService in package.json

CompuIves:


CGNonofr:
Oh yeah, I automated it in some open PRs...

it's probably not the only place where it needs to be added

CompuIves:
I think this should be all places

CGNonofr (approved):

PR #150, opened 2023-07-19
feat(services): Expose view services

DeMoorJasper:
This PR exposes more parts of the view api of vscode, allowing users to build their own services related to view. Like a custom status bar

### DIFF
--- a/src/service-override/views.ts
+++ b/src/service-override/views.ts
@@ -420,5 +420,12 @@ export {
   OpenEditor,
   IEditorOptions,
   IResolvedTextEditorModel,
-  IReference
+  IReference,
+
+  HoverService,
+  ActivityService,
+  StatusbarPart,
+  SidebarPart,
+  ActivitybarPart,
+  PanelPart
 }
--- a/src/services.ts
+++ b/src/services.ts
@@ -65,6 +65,8 @@ export async function initialize (overrides: IEditorOverrideServices): Promise<v
   })
 }
 
+export { SyncDescriptor } from 'vs/platform/instantiation/common/descriptors'
+
 // Export all services as monaco doesn't export them
 export { StandaloneServices } from 'vs/editor/standalone/browser/standaloneServices'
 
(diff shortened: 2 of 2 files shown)


src/service-override/views.ts
```diff
@@ -420,5 +420,20 @@ export {
...
   IEditorOptions,
   IResolvedTextEditorModel,
-  IReference
+  IReference,
+
+  IViewDescriptorService,
+  IActivityService,
+  IHoverService,
+  IExplorerService,
+  IStatusbarService,
```
CGNonofr, line 429: The convention for the moment is to export service identifiers from services.ts

src/service-override/views.ts
```diff
@@ -420,5 +420,20 @@ export {
...
+
+  ViewsService,
+  ViewDescriptorService,
+  ActivityService,
+  HoverService,
+  ExplorerService,
+  StatusbarPart,
+
+  SyncDescriptor
```
CGNonofr, line 438: Probably better to export it from services.ts as well

CGNonofr (changes requested):

src/service-override/views.ts
DeMoorJasper: Alright, changed it so everythings exported from the services file

CGNonofr:
Can you please rewrite your commit names so they follow the conventional commit standard?

src/services.ts
```diff
@@ -209,6 +209,16 @@ export {
   IPromptWithDefaultCancel
 } from 'vs/platform/dialogs/common/dialogs'
 export { IMarkdownString, MarkdownStringTrustedOptions } from 'vs/base/common/htmlContent'
+export { ViewsService } from 'vs/workbench/browser/parts/views/viewsService'
+export { IActivityService, IBadge } from 'vs/workbench/services/activity/common/activity'
+export { ActivityService } from 'vs/workbench/services/activity/browser/activityService'
+export { SyncDescriptor } from 'vs/platform/instantiation/common/descriptors'
+export { IHoverService } from 'vs/workbench/services/hover/browser/hover'
+export { HoverService } from 'vs/workbench/services/hover/browser/hoverService'
```
CGNonofr, line 217: This one is not a service descriptor (They all start with a `I`)

DeMoorJasper:
Addressed all requested changes

CGNonofr:
> Addressed all requested changes

Didn't you need the implementations at the end?

DeMoorJasper:
Ideally yes, but not sure where to export them anymore as you said to only export service descriptors in services

CGNonofr:
> Ideally yes, but not sure where to export them anymore as you said to only export service descriptors in services

Service identifiers should be exported from extensions.ts as you may want to override it without using the vscode implementations and you don't want to pull useless code

Service implementations can be exported from the service override itself without issue

CGNonofr (approved):

PR #153, opened 2023-07-27
feat: add ability to register file getter for local process extensions

nihonium-cfx:
Adds a new `registerFileGetter` function in the `RegisterLocalProcessExtensionResult`.

In my use case I'm bundling custom theme json within the app code and adding local process extension for that, so making it a separate file in the built app is rather an overkill.

In `~1.76.0` I was using the `registerSyncFile` in an object returned by `registerExtension`, but now we've transitioned to different extension hosts, so I understand the need in API change, just that for local process exthost this new API function should not be a problem.

### DIFF
--- a/src/extensions.ts
+++ b/src/extensions.ts
@@ -31,6 +31,7 @@ interface RegisterExtensionResult {
 interface RegisterLocalProcessExtensionResult extends RegisterExtensionResult {
   getApi (): Promise<typeof vscode>
   setAsDefaultApi (): Promise<void>
+  registerFileGetter: (path: string, getter: () => Promise<Uint8Array>) => IDisposable
 }
 
 function registerExtensionFileUrl (extensionLocation: URI, filePath: string, url: string, mimeType?: string): IDisposable {
@@ -116,6 +117,9 @@ export function registerExtension (manifest: IExtensionManifest, extHostKind?: E
       getApi,
       async setAsDefaultApi () {
         setDefaultApi(await getApi())
+      },
+      registerFileGetter (path, getter) {
+        return registerExtensionFile(extension.location, path, getter)
       }
     }
   }


CGNonofr:
Thanks for your contribution!

The issue with the suggested solution is it won't work for every files (for instance css files).

If I understand your need, you want to inline a file content in your code (or generate it on the fly?)

You can still do it using a data url, you can use this function:

```javascript
async function stringToDataUrl (content: string, mimeType: string): Promise<string> {
  return new Promise<string>((resolve) => {
    const reader = new FileReader()
    reader.onloadend = function () {
      resolve(reader.result as string)
    }
    reader.readAsDataURL(new Blob([content], { type: mimeType }))
  })
}
```

note that it's probably a better idea to put that content in its own file, and let your bundler decide if it wants to inline it or not

nihonium-cfx:
> The issue with the suggested solution is it won't work for every files (for instance css files).

Why would it not work for css files? If it's a main-thread running code and no headers or other metadata retained in the getter for URLs, I don't really understand why would it not work.

> You can still do it using a data url

> note that it's probably a better idea to put that content in its own file, and let your bundler decide if it wants to inline it or not

I understand where you're coming from here and normally I'd give such an advise as well, but that's not how I need my IDE to operate and I'm trying to optimize latency to loaded theme here.

There's another thing I'm exploring right now and  that's a theme colors cache similar to how vscode does it, so editor does not flash from default theme to user-selected one on load, but that's for another PR 😛

CGNonofr:
> Why would it not work for css files? If it's a main-thread running code and no headers or other metadata retained in the getter for URLs, I don't really understand why would it not work.

Because vscode transforms the extension url to a browser url (using `FileAccess` class) and just injects the url in the dom. AFAIK there is no way to fetch the data at this point since the method is synchronous so we need to have an accessible url beforehand.
It may work for your use-case, but adding it in the api may be confusing for someone with another use-case. (A lot of stuff in an extension wouldn't work, at least everything that was marked async in https://github.com/CodinGame/monaco-vscode-api/blob/v1.79.13/src/extension-tools.ts: all language/command/views icons, theme fonts, jsonValidation files, css files...).
Is there an issue with the suggestion solution?

> There's another thing I'm exploring right now and that's a theme colors cache similar to how vscode does it, so editor does not flash from default theme to user-selected one on load, but that's for another PR 😛

Exciting! can't wait to have a look at it!

CGNonofr:
Always a pleasure!

PR #154, opened 2023-07-27
feat: add overridable storage service

nihonium-cfx:
A followup for https://github.com/CodinGame/monaco-vscode-api/pull/153#issuecomment-1653683173

With this new override one can provide monaco with data for it's storage service, which is, among others, responsible for storing cached theme data. This allows for monaco to restore theme colors without flickering through default theme first, if you supply the storage data it asked you store before, of course.

Example usage:
```typescript
// The settings.json file needs to be written ASAP
registerServiceInitializeParticipant(async (accessor) => {
  // import { writeFile } from 'vscode/monaco'
  await writeFile(URI.from({ schema: 'user', path: '/settings.json' }), `
  {
    "workbench.colorTheme": "Monokai",
  }
  `);
});

await initialize({
  ...getStorageServiceOverride({
    read(scope) {
      try {
        const data = JSON.parse(localStorage[`__STORAGE__${scope}__`]);

        return new Map(data);
      } catch (e) {
        return new Map();
      }
    },
    async write(scope, data) {
      localStorage[`__STORAGE__${scope}__`] = JSON.stringify(Array.from(data));
    }
  }),
});
```

### DIFF
--- a/src/service-override/storage.ts
+++ b/src/service-override/storage.ts
@@ -1,0 +1,128 @@
+import { Event } from 'vs/base/common/event'
+import { IDisposable } from 'vs/base/common/lifecycle'
+import { IStorage, IStorageDatabase, IStorageItemsChangeEvent, IUpdateRequest, Storage } from 'vs/base/parts/storage/common/storage'
+import { IEditorOverrideServices } from 'vs/editor/standalone/browser/standaloneServices'
+import { SyncDescriptor } from 'vs/platform/instantiation/common/descriptors'
+import { AbstractStorageService, IStorageService, StorageScope as VSStorageScope } from 'vs/platform/storage/common/storage'
+import { IUserDataProfile } from 'vs/platform/userDataProfile/common/userDataProfile'
+import { IAnyWorkspaceIdentifier } from 'vs/platform/workspace/common/workspace'
+
+export enum StorageScope {
+  APPLICATION = VSStorageScope.APPLICATION,
+  PROFILE = VSStorageScope.PROFILE,
+  WORKSPACE = VSStorageScope.WORKSPACE
+}
+
+export interface IStorageProvider {
+  read (scope: StorageScope): Map<string, string> | undefined
+  write (scope: StorageScope, data: Map<string, string>): Promise<void>
+  close? (scope: StorageScope): Promise<void>
+  onDidChange? (listener: (event: IStorageItemsChangeEvent) => void): IDisposable
+}
+
+class ExternalStorage extends Storage {
+  constructor (
+    scope: StorageScope,
+    provider: IStorageProvider
+  ) {
+    const items = provider.read(scope)
+
+    super(new ExternalStorageDatabase(scope, provider, items))
+
+    if (items != null) {
+      for (const [key, value] of items) {
+        this.items.set(key, value)
+      }
+    }
+  }
+}
+
+class ExternalStorageDatabase implements IStorageDatabase {
+  readonly onDidChangeItemsExternal: Event<IStorageItemsChangeEvent>
+
+  constructor (
+    private readonly scope: StorageScope,
+    private readonly provider: IStorageProvider,
+    private readonly items = new Map<string, string>()
+  ) {
+    this.onDidChangeItemsExternal = this.provider.onDidChange ?? Event.None
+  }
+
+  async getItems (): Promise<Map<string, string>> {
+    return this.items
+  }
+
+  async updateItems (request: IUpdateRequest): Promise<void> {
+    request.insert?.forEach((value, key) => this.items.set(key, value))
+
+    request.delete?.forEach(key => this.items.delete(key))
+
+    await this.provider.write(this.scope, this.items)
+  }
+
+  async close () {
+    return this.provider.close?.(this.scope)
+  }
+}
+
+class ExternalStorageService extends AbstractStorageService {
+  private readonly applicationStorage = this._register(new ExternalStorage(StorageScope.APPLICATION, this.provider))
+  private readonly profileStorage = this._register(new ExternalStorage(StorageScope.PROFILE, this.provider))
+  private readonly workspaceStorage = this._register(new ExternalStorage(StorageScope.WORKSPACE, this.provider))
+
+  constructor (protected readonly provider: IStorageProvider) {
+    super()
+
+    this._register(this.workspaceStorage.onDidChangeStorage(key => this.emitDidChangeValue(VSStorageScope.WORKSPACE, key)))
+    this._register(this.profileStorage.onDidChangeStorage(key => this.emitDidChangeValue(VSStorageScope.PROFILE, key)))
+    this._register(this.applicationStorage.onDidChangeStorage(key => this.emitDidChangeValue(VSStorageScope.APPLICATION, key)))
+  }
+
+  protected getStorage (scope: VSStorageScope): IStorage {
+    switch (scope) {
+      case VSStorageScope.APPLICATION:
+        return this.applicationStorage
+      case VSStorageScope.PROFILE:
+        return this.profileStorage
+      default:
+        return this.workspaceStorage
+    }
+  }
+
+  protected getLogDetails (scope: VSStorageScope): string | undefined {
+    switch (scope) {
+      case VSStorageScope.APPLICATION:
+        return 'External (application)'
+      case VSStorageScope.PROFILE:
+        return 'External (profile)'
+      default:
+        return 'External (workspace)'
+    }
+  }
+
+  protected async doInitialize (): Promise<void> {
+    // no-op
+  }
+
+  protected async switchToProfile (): Promise<void> {
+    // no-op
+  }
+
+  protected async switchToWorkspace (): Promise<void> {
+    // no-op
+  }
+
+  hasScope (_scope: IAnyWorkspaceIdentifier | IUserDataProfile): boolean {
+    return false
+  }
+}
+
+export default function getStorageServiceOverride (provider: IStorageProvider): IEditorOverrideServices {
+  return {
+    [IStorageService.toString()]: new SyncDescriptor(ExternalStorageService, [provider])
+  }
+}
+
+export {
+  IStorageItemsChangeEvent
+}


src/service-override/storage.ts
```diff
@@ -0,0 +1,144 @@
+import { Emitter, Event } from 'vs/base/common/event'
+import { IStorage, IStorageDatabase, IStorageItemsChangeEvent, IUpdateRequest, Storage } from 'vs/base/parts/storage/common/storage'
+import { IEditorOverrideServices } from 'vs/editor/standalone/browser/standaloneServices'
+import { SyncDescriptor } from 'vs/platform/instantiation/common/descriptors'
+import { AbstractStorageService, IStorageService, StorageScope as VSStorageScope } from 'vs/platform/storage/common/storage'
+import { IUserDataProfile } from 'vs/platform/userDataProfile/common/userDataProfile'
+import { IAnyWorkspaceIdentifier } from 'vs/platform/workspace/common/workspace'
+
+export enum StorageScope {
```
nihonium-cfx, line 9: Original `StorageScope` is a `const enum` so can not be properly reexported unless we remove `const` in patch, figured that's more convenient.

src/service-override/storage.ts
```diff
@@ -0,0 +1,144 @@
...
+  APPLICATION = 'application',
+  PROFILE = 'profile',
+  WORKSPACE = 'user',
+}
+
+export interface IStorageProvider {
+  read (scope: StorageScope): Map<string, string> | undefined
+  write (scope: StorageScope, data: Map<string, string>): Promise<void>
```
nihonium-cfx, line 17: Now that I'm rereading the code, I think the return type of `read` and second arg of `write` should be something that can already be serialized easily by provider w/o the need to deal with `Map`s.

src/service-override/storage.ts
CGNonofr: Yeah, vscode uses a lot of const enum. the build process here preserves them but the ones in monaco-editor are not preserved. I don't have any perfect solution but I've already had the same issue for another enum and I've overriden it in the `override` folder (see https://github.com/CodinGame/monaco-vscode-api/blob/8c0e697dfc8c599dc7b3499adda6122b19050f5c/src/override/vs/platform/dialogs/common/dialogs.ts)

src/service-override/storage.ts
```diff
@@ -0,0 +1,144 @@
...
+  ) {
+    if (this.provider.onDidChange != null) {
+      const emitter = new Emitter<IStorageItemsChangeEvent>()
+
+      this.provider.onDidChange((data: IStorageItemsChangeEvent) => emitter.fire(data))
+
+      this.onDidChangeItemsExternal = emitter.event
+    } else {
+      this.onDidChangeItemsExternal = Event.None
+    }
```
CGNonofr, line 55: Isn't it equivalent to `this.onDidChangeItemsExternal = this.provider.onDidChange ?? Event.None` ?

src/service-override/theme.ts
```diff
@@ -33,7 +33,7 @@ type PartialIThemeExtensionPoint = Partial<IThemeExtensionPoint> & Pick<IThemeEx
 export default function getServiceOverride (): IEditorOverrideServices {
   return {
     ...getFileServiceOverride(),
-    [IThemeService.toString()]: new SyncDescriptor(StandaloneWorkbenchThemeService)
+    [IThemeService.toString()]: new SyncDescriptor(StandaloneWorkbenchThemeService, undefined, true)
```
CGNonofr, line 36: You said you want it to be eager but the last parameter seems to do the opposite: it indicate if the service supports delayed instantiation

src/service-override/storage.ts
nihonium-cfx: Ultimately yes, that's the equivalent, however signatures do not match, but if you reckon it's better for `provider.onDidChange` to match the `Event` signature instead - will change to that!

src/service-override/theme.ts
nihonium-cfx: I, of course, meant the delayed instantiation. So we want this service to be instantiated not when it is resolved as a dependency of another service, but when it is actually used, this way we have a bit more time to write the settings file so theme id from settings would be picked up and all the logic loading the cached theme data would work.

It still feels sloppy this implementation and I'd prefer not counting on that we will most likely write settings file timely, but that's a start!

src/service-override/storage.ts
nihonium-cfx: My reasoning to make it a string literal enum was that for implementer it removes the need to convert to strings as ultimately you would be just writing and reading these data objects and it's nice to have a string for scope already as I'd bet usually one wants a string as a name for these object keys. But yeah, if you prefer overriding like with the `ConfirmResult` - I'll change to that, just let me know

src/service-override/theme.ts
nihonium-cfx: Okay, I see the problem now, in vscode it indeed is eager and not delayed. Thing is, when I was testing cached values were not picked up and it was still flickering because `settings.json` was not yet written so changing it to delayed did help a ton and so far have been working fine. But as said, that's really a very reliable solution

nihonium-cfx: The main problem here is that we do not have a configuration service initialized with the data from provided `user:/settings.json` as our service init sequence does not wait for that to be done.

I think the best solution for that might be a way to provide all the configuration data before initializing services and teach our config service override to use that data instead of relying on read that from fileService.

Otherwise we'd need to implement custom services init sequence where we'd wait for certain services to async init first, which I presume is undesirable

CGNonofr: It seems the storage service needs to be initialized. to do so, you call use `registerServiceInitializeParticipant`

Then you can implement `doInitialize` and do whatever you need in there, right?

nihonium-cfx: > It seems the storage service needs to be initialized. to do so, you call use `registerServiceInitializeParticipant`
> 
> Then you can implement `doInitialize` and do whatever you need in there, right?

Let me expand on this:

In order to avoid theme flickering both these requirements must be met, ideally:
1. Storage data to be available **during** the construction of `WorkspaceThemeService`, in our case it's during `initialize()` call
2. `workbench.colorTheme` configuration value to also be available **during** the construction of `WorkspaceThemeService`

First requirement is gonna be covered once this PR is merged, but the second one is currently impossible to do because we can only write the configuration file content (i.e. making `workbench.colorTheme` value be readable from the configuration service) **after** services has been initialized. There's no way to make the content of `user:/settings.json` available to configuration service prior to initialization of `WorkspaceThemeService`.

And so right now the weak workaround is to make `WorkspaceThemeService` to support delayed instantiation so we have a chance to write settings file **after** initialization and **before** actual construction of `WorkspaceThemeService`.

Hope this explains, haha.

Right now im just passing my own configuration service as an override so I can guarantee availability of all needed user settings during the monaco services initialization, but this is far from ideal because, well, I literally just copied config service code straight from vscode to ensure that monaco can use that w/o any incompatibilities.

src/service-override/storage.ts
nihonium-cfx: > So will you change something here at the end?

No, in the end I don't really think it's that big of a deal to work with maps for the implementer of provider.

src/service-override/storage.ts
nihonium-cfx: > I understand your point, but doing it for a single enum is not very consistent. Also I think we better modify vscode code as less as possible, we don't really know what is done with that enum. There is a negative value, maybe there is a reason for it, maybe the enum value is compared somewhere, maybe it'll change in the future, who knows 🤷

Oh yes, absolutely, vscode patching must as minimal as possible. Learned that the hard way trying to maintain the fork of vscode for our needs. And that's why we're currently migrating away from vscode fork.

I will change this enum to a non-const copy of original one.

src/service-override/theme.ts
CGNonofr: Looking at web.main.ts, it seems that workspace/configuration and storage services are initialized first, then all other services. We probably need something similar here

nihonium-cfx: > Looking at web.main.ts, it seems that workspace/configuration and storage services are initialized first, then all other services. We probably need something similar here

Yes, that's exactly what I meant, that ideally we'd want to migrate to init strategy like that and away from how standalone services do it.

That can be done in future PRs and I don't think that should be a blocker for this. How do you think?

src/service-override/storage.ts
nihonium-cfx: > I'm not really sure to understand what you mean by
> 
> > signatures do not match

Well the difference is that `provider.onDidChange` return type is `void` and `Event`'s return type is `IDisposable`. I can either change provider to use `IDisposable` or keep the current solution.

Reasoning was that forcing implementer to return disposable is a bit too much as it most likely won't be called at all.

src/service-override/theme.ts
CGNonofr: If I understand correctly, this PR is a partial fix and you still need to hack to fulfill your needs?

nihonium-cfx: > If I understand correctly, this PR is a partial fix and you still need to hack to fulfill your needs?

It is, indeed, partial right now.

Okay, let me just remove this theme service delayed instantiation as it is just a bad solution anyway.

> A proper fix seems quite simple: we probably just need to have a `registerServiceInitializeParticipant` AND a `registerServicePreInitializeParticipant` or something similar, and use the second one for configuration and storage

Ideally, yes, configuration service being inited prior to everything else will fix the issue in it's fullness. With this PR you don't need to preinit storage tho as provider here provides data synchronously so you would wait for that data prior to services initialization anyway.

nihonium-cfx:
I think it's better to address init sequence concerns in another PR

src/service-override/storage.ts
```diff
@@ -0,0 +1,146 @@
...
+}
+
+export function StorageScopeToString (scope: StorageScope): string {
+  switch (scope) {
+    case StorageScope.APPLICATION: return 'APPLICATION'
+    case StorageScope.PROFILE: return 'PROFILE'
+    case StorageScope.WORKSPACE: return 'WORKSPACE'
+    default: return `UNKNOWN.${scope}`
+  }
+}
```
CGNonofr, line 23: is it used somewhere?

src/service-override/storage.ts
```diff
@@ -0,0 +1,146 @@
...
+const defaultStorageProvider: IStorageProvider = {
+  read () {
+    return new Map()
+  },
+  async write () {
+
+  }
+}
+
+export default function getStorageServiceOverride (provider = defaultStorageProvider): IEditorOverrideServices {
```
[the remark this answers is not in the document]
nihonium-cfx: No, not really. Is it about having a default no-op provider?

src/service-override/storage.ts
nihonium-cfx: That's a helper for provider implementer. Like in my case I don't really care about the actual scope and just need some string identifier to store it.

src/service-override/storage.ts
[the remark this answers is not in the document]
CGNonofr: yep, not a big deal though

src/service-override/storage.ts
CGNonofr: I think I prefer to keep that user implementation specific code in the user code and not in that library

PR #158, opened 2023-08-04
Always register file service override

### LINKED ISSUE #156: Initialization failing
I'm currently testing combo of `monaco-editor@0.40.0` and `@codingame/monaco-vscode-api@1.80.0`. It looks like neither works as I expect it to work. Maybe I'm doing something wrong. Here are the simplest POC:

### Initializing services

```js
import { initialize as initializeMonacoServices } from 'vscode/services';

await initializeMonacoServices({});
```

This is the result in browser console:

image

Alternatively I can use following code to initialize standalone services:

```js
import { StandaloneServices } from 'vscode/services';

StandaloneServices.initialize({});
```

This works without errors (as it have been working in previous versions of `@codingame/monaco-vscode-api`, but not sure how it now play with initializing the `extensions` below.

### Initializing extensions

```js
import { initialize as initializeVscodeExtensions } from 'vscode/extensions';

await initializeVscodeExtensions();
```
`initializeVscodeExtensions` always returns promise that never fulfills. I can see  that the code in https://github.com/CodinGame/monaco-vscode-api/blob/main/demo/src/setup.ts seems to work for this repo `demo` but I have no idea why simplest imports wouldn't work now.

CGNonofr:
fix #156

### DIFF
--- a/src/services.ts
+++ b/src/services.ts
@@ -14,6 +14,7 @@ import { IAction } from 'vs/base/common/actions'
 import getLayoutServiceOverride from './service-override/layout'
 import getEnvironmentServiceOverride from './service-override/environment'
 import getExtensionsServiceOverride from './service-override/extensions'
+import getFileServiceOverride from './service-override/files'
 
 interface ServiceInitializeParticipant {
   (accessor: ServicesAccessor): Promise<void>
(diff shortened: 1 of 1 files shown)


kaisalmen (approved):
LGTM. Then there is no need to even configure it / allow to configure it with monaco-languageclient

Issue #162, opened 2023-08-14
Uncaught Error: Cannot have two contributions with the same id editor.contrib.referencesController


Seventysevendays:
initialize got error, what's the problem
![image](https://github.com/CodinGame/monaco-vscode-api/assets/37206679/4354d65f-b294-40cd-8b59-c993e8b5cd4b)

CGNonofr:
This library pulls back vscode features into monaco editor

monaco-editor is trying to do the same with standalone versions of the features, and it conflicts

The solution is to never import `monaco-editor` but `monaco-editor/esm/vs/editor/editor.api.js` instead (you can use your bundler alias feature to do it)

for more details, [these files](https://github.com/microsoft/vscode/blob/b7ebbb6990b1d2a04b5cbbd816cfa73f96b1c1d8/src/vs/editor/editor.main.ts#L8-L14) are specific to the standalone editor (monaco) while there is a vscode version of it (pulled here)

you should probably keep the other imports though ( `monaco-editor/esm/vs/editor/editor.all.js` to keep editor-related features and `vs/editor/standalone/browser/iPadShowKeyboard/iPadShowKeyboard` which is only used in monaco)

Seventysevendays:
> This library pulls back vscode features into monaco editor
> 
> monaco-editor is trying to do the same with standalone versions of the features, and it conflicts
> 
> The solution is to never import `monaco-editor` but `monaco-editor/esm/vs/editor/editor.api.js` instead (you can use your bundler alias feature to do it)
> 
> for more details, [these files](https://github.com/microsoft/vscode/blob/b7ebbb6990b1d2a04b5cbbd816cfa73f96b1c1d8/src/vs/editor/editor.main.ts#L8-L14) are specific to the standalone editor (monaco) while there is a vscode version of it (pulled here)
> 
> you should probably keep the other imports though ( `monaco-editor/esm/vs/editor/editor.all.js` to keep editor-related features and `vs/editor/standalone/browser/iPadShowKeyboard/iPadShowKeyboard` which is only used in monaco)

it works, thanks so much

jhk-mjolner:
For anyone finding themselves here, wondering if they should import monaco-editor as `monaco-editor/esm/vs/editor/editor.api.js` for whatever reason:

You should not. According to https://github.com/TypeFox/monaco-languageclient/issues/562#issuecomment-1776796703

It can cause bugs to appear, such as causing the message for an editor set to be readonly to not appear.
[...]

PR #163, opened 2023-08-14
Fix 1.81 introduced issues

CGNonofr:
What it does:
- do not require to import only `editor.api` by patching monaco files and removing monaco specific features
- demo storage service (view locations...)
- fix search view
- improve view service override: directly create parts in invisible elements ; it fixes the usage of the storage service for the view locations ; and doesn't require the user to render the part to prevent crashing
- it fixes the typescript-language-feature extension that was not able anymore to load the require files


kaisalmen:
> do not require to import only editor.api by patching monaco files and removing monaco specific features

But then the behaviour is different, because you have to load the services in addition to have the same features. This is desired, because everything is available here now, but this must be clear to the end-user.

Was just wondering if it makes sense to have function that does exactly that: Import monaco and configure/return all vscode version of the services that you need to have the same as `import * as monaco from 'monaco-editor'` from the unaltered `monaco-editor`.

Can you release a preview version? Thanks

CGNonofr:
> > do not require to import only editor.api by patching monaco files and removing monaco specific features
> 
> @CGNonofr but then the behaviour is different, because you have to load the services in addition to have the same features. This is desired, because everything is available here now, but this must be clear to the end-user.

(it's related to https://github.com/CodinGame/monaco-vscode-api/issues/162)

true for the quickaccess, others things are imported directly in https://github.com/CodinGame/monaco-vscode-api/blob/3b1acfe0e1028b744e3d5f19785d34cef107f3d4/src/monaco.ts#L45-L56

The bundler configuration to remove monaco features really seems confusing for users though

> Was just wondering if it makes sense to have function that does exactly that: Import monaco and configure/return all vscode version of the services that you need to have the same as `import * as monaco from 'monaco-editor'` from the unaltered `monaco-editor`. WDYT?

I'm not sure to understand what you exactly mean, can you explain a little bit more?

> Can you release a preview version? Thanks

I forgot every time, maybe I should automate it => @codingame/monaco-vscode-api@1.81.1-next.0

kaisalmen:
@CGNonofr just thought abut it again, maybe it is better to pass in the services you want and then add what is needed and perform `initializeMonacoService` inside that function.

CGNonofr:
@kaisalmen we already have default service override => https://github.com/CodinGame/monaco-vscode-api/blob/3b1acfe0e1028b744e3d5f19785d34cef107f3d4/src/services.ts#L28-L34

Do you suggest to add quickAccess in here?

kaisalmen:
> Do you suggest to add quickAccess in here?

Why not there, yes. It is such a useful feature. I usually configure it everywhere usually.

CGNonofr:
> > Do you suggest to add quickAccess in here?
> 
> Why not there, yes. It is such a useful feature. I usually configure it everywhere usually.

How to handle the getServiceOverride parameters then?

kaisalmen:
> How to handle the getServiceOverride parameters then?

Configure it differently if views service is used part of the overrides?

CGNonofr:
> > How to handle the getServiceOverride parameters then?
> 
> Configure it differently if views service is used part of the overrides?

How can we know?

kaisalmen:
> How can we know?

Check the overrides like I do in monaco-langaugeclient. Reminds me that I still have to check the new next build

kaisalmen (commented):
Integration & tests with monaco-languageclient were successful.

CGNonofr:
@kaisalmen what do you think about 764200d6ace51af9c4e742434eb4f4384f45e968 ?

published as @codingame/monaco-vscode-api@1.81.1-next.1

kaisalmen:
@CGNonofr so if I understand correctly, then when views service is used it overrides the default quickAccess service

CGNonofr:
> @CGNonofr so if I understand correctly, then when views service is used it overrides the default quickAccess service

It overrides it no matter what, but it over-overrides it when using the view service to inject the parameters

kaisalmen (approved):
LGTM. Tested successfully here: https://github.com/TypeFox/monaco-languageclient/pull/523

PR #165, opened 2023-08-18
Service description

kaisalmen:
Once a start is made, continuation is simpler 🙂 It is not rocket science, but I hope you can complement.

Specific questions:
- Editor: Why is not required by default? What is the difference to not configuring it at all
- Languages: Is this a must if for LSP?

What do these services actually do?
- Language detection worker
- Dialogs
- Snippets

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,25 +123,29 @@ await initialize({
 Additionally, this library exposes 23 modules that include the vscode version of some services (with some glue to make it work with monaco):
 
 - **Extensions** (included by default): `vscode/service-override/extensions`
-  - Support for VSCode extensions.
+  - Support for VSCode extensions. A worker configuration can be past to it:
+    - Then, everything runs in one worker, where extensions run in an iframe, with all the implications (can be created by the bundler directly). The worker script is expected to be hosted on a separate domain.
 - **Files** (included by default): `vscode/service-override/files`
-  - It adds the memory filesystem for `file://` files, but also adds the support for lazy loaded extension files, separate memory user files (e.g. config, keybindings), cache, logs, etc.
+  - It adds the memory filesystem for `file://` files, but also adds the support for lazy loaded extension files. It adds separate memory user files (e.g. config, keybindings), cache files and log files.
 - **QuickAccess** (included by default): `vscode/service-override/quickaccess`
-  - Enables the quickaccess menu in the editor
+  - Enables the quickaccess menu in the editor (press F1)
 - **Notifications**: `vscode/service-override/notifications`
-  - This services enables vscode notifications (in the bottom right corner)
+  - This services enables vscode notifications you usually find in the bottom right corner.
 - **Dialogs**: `vscode/service-override/dialogs`
-  - Enable vscode modal dialogs. They allow users to select an action to perform which are exposed to the vscode API. Additionally, this can be used by the language client to ask questions to the user?
+  - Enable vscode modal dialogs. It allows users to select actions to perform and exposed them the vscode API. Additionally, this service can be used by the language client to delegate questions to the user.
 - **Model**: `vscode/service-override/model`
-  - This service creates and takes care of model references (e.g. create model if unknown, count references and destroy model when not used anymore)
+  - This service creates and takes care of model references. For example:
+    - Create model if content is unknown
+    - Count references
+    - Destroy models when they are no lonbger used
 - **Editor**: `vscode/service-override/editor`
-  - Enable editor support. This is usually not needed when working without LSP. Without enabling the editor service, it will only be able to resolve the currently open model, so only internal file links will work.
+  - Enable editor support. This is usually needed when working with the language server protocol. Without enabling the editor service, it will only be able to resolve the currently open model (only internal file links will work).
 - **Views**: `vscode/service-override/views`
-  - Enable full views support. Is exclusive with `editor`, do not use both at the same time.
+  - Enable full views support. Is exclusive with the `editor` service. Do not use both services at the same time.
 - **Configuration**: `vscode/service-override/configuration`
   - Allows to change the configuration of not only the editors, but every part of vscode. The language client for instance uses it to send the requested configuration to the server. The default configuration service already allows to change the configuration. This service overrides makes it rely on a user configuration file (with json schema, overridable by language including all vscode features).
 - **Keybindings**: `vscode/service-override/keybindings`
-  - Enables platform specific keybindings and make it rely on a user definded keybindings configuration (if available)
+  - Enables platform specific keybindings and make it rely on a user definded keybindings configuration (if available).
 - **Languages**: `vscode/service-override/languages`
   - Enable language support. It's like the standalone service with 2 differences:
     - It handle the language extension point (getting languages from vscode extensions)
--- a/README.md
+++ b/README.md
@@ -122,55 +122,61 @@ await initialize({
 
 Additionally, this library exposes 23 modules that include the vscode version of some services (with some glue to make it work with monaco):
 
-- Extensions (included by default): `vscode/service-override/extensions`
-  - Support for VSCode extensions
-- Files (included by default): `vscode/service-override/files`
-  - File system support
-- QuickAccess (included by default): `vscode/service-override/quickaccess`
+- **Extensions** (included by default): `vscode/service-override/extensions`
+  - Support for VSCode extensions.
+- **Files** (included by default): `vscode/service-override/files`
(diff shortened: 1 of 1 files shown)


CGNonofr:
> Editor: Why is not required by default? What is the difference to not configuring it at all

> TODO: Why is not enabled by default?

It's not enabled by default because:
- it's not absolutely required: it works without it, you just won't get redirected to another model. It's typically not required without LSP
- it conflicts with the views service override

without it, it will only be able to resolve the currently open model, so only internal file links will work

> Languages: Is this a must if for LSP?

Languages need to be registered or the language client language selector won't work, but it can either be done by vscode extensions or monaco monarch languages

To have a fully functional LSP, you'll need semantic tokens, which require textmate grammars, with require vscode extensions which also include related languages, so it don't really see the point of not including it

## in the code

> - Files (included by default): `vscode/service-override/files`
>   - File system support

Note: it adds the memory filesystem for `file://` files, but also adds the support for lazy loaded extension files, separate memory user files (config, keybindings...), cache, logs...

> - Notifications: `vscode/service-override/notifications`
>   - Enables sending and reciving notifications

Not sure what you mean by `receiving` notifications: it includes the vscode notifications (in the bottom right corner)

> - Dialogs: `vscode/service-override/dialogs`
>   - ?

It's the vscode modal/dialogs. They allow the user to select an action to do and are exposed to the vscode api AND can be used by the language client to ask questions to the user

> - Model: `vscode/service-override/model`
>   - Handle text models

That's the service that creates and takes care of model references (create model if unknown, count references, destroy model when not used anymore...)

> - Configuration: `vscode/service-override/configuration`
>   - Alllows to change the configuration of the editor during runtime 

Not only the editors, but every part of vscode. The language client for instance uses it to send the requested configuration to the server.

The default configuration service already allows to change the configuration

This service overrides makes it rely on a user configuration file (with json schema, overridable by language... all vscode features)

> - Keybindings: `vscode/service-override/keybindings`
>   - Enables platform specific keybindings

It also make the keybindings rely on a user keybindings configuration files

> - Languages: `vscode/service-override/languages`
>   - Enable language support.

It's like the standalone service with 2 differences:
- it handle the language extension point (getting languages from vscode extensions
- it triggers the `onLanguage:${language}` event (to load vscode extension listening of those events)

> - Textmate: `vscode/service-override/textmate`
>   - Allows to use textmate grammars. Depends on theme service.

Maybe it's worth it to add that vscode extensions use textmate grammars exclusively for highlighting

> - Snippets: `vscode/service-override/snippets`
>   - ?

Add snippet extension point (register vscode extension snippets)

> - Output: `vscode/service-override/output`
>   - Output panel support
> - Terminal: `vscode/service-override/terminal`
>   - Terminal panel support
> - Search: `vscode/service-override/search`
>   - search panel support
> - Markers: `vscode/service-override/markers`

It may be relevant to add that those service overrides make sense only if the view service override is used

(the marker service override adds the `problems` panel tab)

> - Language detection worker: `vscode/service-override/languageDetectionWorker`
>   - ?

(It's a quite recent adding here)

When opening an untitled model or a file without extension or if vscode is unable to guess the language simply by the file extension or by reading the first line, it will use tensorflow in a worker to try to guess the most probable language

Note that in vscode, they used 2 models, 1 open and 1 closed source, we're only able to use the open one here

> - Storage: `vscode/service-override/storage`
>   - Define your own storage

Storage service is used at a lot of places either as a cache or as a user preference store. For instance:
- Current loaded theme is stored in there to be loaded faster on start
- every panel/view position are stored in there

kaisalmen:
@CGNonofr wow, these are very extensive answers. Thanks. Now, I need to incorporate them...

kaisalmen:
@CGNonofr I incorporated your comments. Please check if there are errors or if there is need for improvement.

Should we add something regarding extension workers?

README.md
```diff
@@ -122,33 +122,61 @@ await initialize({
...
+- **Extensions** (included by default): `vscode/service-override/extensions`
+  - Support for VSCode extensions.
+- **Files** (included by default): `vscode/service-override/files`
+  - It adds the memory filesystem for `file://` files, but also adds the support for lazy loaded extension files, separate memory user files (e.g. config, keybindings), cache, logs, etc.
+- **QuickAccess** (included by default): `vscode/service-override/quickaccess`
+  - Enables the quickaccess menu in the editor
+- **Notifications**: `vscode/service-override/notifications`
+  - This services enables vscode notifications (in the bottom right corner)
+- **Dialogs**: `vscode/service-override/dialogs`
+  - Enable vscode modal dialogs. They allow users to select an action to perform which are exposed to the vscode API. Additionally, this can be used by the language client to ask questions to the user?
```
CGNonofr, line 134: It is a question? :smile:

CGNonofr:
> @CGNonofr I incorporated your comments. Please check if there are errors or if there is need for improvement.

I was thinking you would reformulate things to make it a little bit more... formal? (:

> Should we add something regarding extension workers?

Probably yeah, there is a lot to say regarding it as well (there is only 1 worker though)

What can be said:
- it run from an iframe, with all the implications (can be created by the bundler directly...)
- It expect the worker script to be hosted on a separate domain

README.md
kaisalmen: I will correct it.

kaisalmen:
> I was thinking you would reformulate things to make it a little bit more... formal? (:

I already did, but should do another round, I guess 🙂

kaisalmen:
@CGNonofr second update is done!

README.md
```diff
@@ -123,25 +123,29 @@ await initialize({
...
 - **Extensions** (included by default): `vscode/service-override/extensions`
-  - Support for VSCode extensions.
+  - Support for VSCode extensions. A worker configuration can be past to it:
+    - Then, everything runs in one worker, where extensions run in an iframe, with all the implications (can be created by the bundler directly). The worker script is expected to be hosted on a separate domain.
 - **Files** (included by default): `vscode/service-override/files`
-  - It adds the memory filesystem for `file://` files, but also adds the support for lazy loaded extension files, separate memory user files (e.g. config, keybindings), cache, logs, etc.
+  - It adds the memory filesystem for `file://` files, but also adds the support for lazy loaded extension files. It adds separate memory user files (e.g. config, keybindings), cache files and log files.
 - **QuickAccess** (included by default): `vscode/service-override/quickaccess`
-  - Enables the quickaccess menu in the editor
+  - Enables the quickaccess menu in the editor (press F1)
```
CGNonofr, line 131: or ctrl+shift+p/ctrl+p like in vscode :)

README.md
```diff
@@ -123,25 +123,29 @@ await initialize({
...
 - **Notifications**: `vscode/service-override/notifications`
-  - This services enables vscode notifications (in the bottom right corner)
+  - This services enables vscode notifications you usually find in the bottom right corner.
 - **Dialogs**: `vscode/service-override/dialogs`
-  - Enable vscode modal dialogs. They allow users to select an action to perform which are exposed to the vscode API. Additionally, this can be used by the language client to ask questions to the user?
+  - Enable vscode modal dialogs. It allows users to select actions to perform and exposed them the vscode API. Additionally, this service can be used by the language client to delegate questions to the user.
```
CGNonofr, line 135: I'm not sure about this sentence:

> It allows users to select actions to perform and exposed them the vscode API.

What is exposed if the possibility to ask something to the user

README.md
```diff
@@ -123,25 +123,29 @@ await initialize({
...
 - **Model**: `vscode/service-override/model`
-  - This service creates and takes care of model references (e.g. create model if unknown, count references and destroy model when not used anymore)
+  - This service creates and takes care of model references. For example:
+    - Create model if content is unknown
+    - Count references
+    - Destroy models when they are no lonbger used
```
CGNonofr, line 140: (small typo)

README.md
kaisalmen: <F1> works in vscode 😉

kaisalmen: see commit ⬇️

README.md
kaisalmen: I took a look again what you initially wrote. What about: 
"It allows users to select an action to do. Those actions are exposed to the vscode API..."

I search for something in the vscode docs, but that didn't lead anywhere. Are you ok with this now? ⬇️

README.md
kaisalmen: Corrected ⬇️

CGNonofr (approved):

kaisalmen:
🎉

PR #168, opened 2023-08-21
Improve view management

CGNonofr:
published as @codingame/monaco-vscode-api@1.81.3-next.0

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -21,7 +21,11 @@
           <button id="filesystem">Attach filesystem</button>
           <button id="customEditorPanel">Open custom editor panel</button>
           <button id="clearStorage">Clear storage</button>
+          <br />
+          <button id="togglePanel">Toggle Panel</button>
+          <button id="toggleAuxiliary">Toggle Secondary Panel</button>
         </div>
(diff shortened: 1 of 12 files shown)


CGNonofr:
> * I see the "Clear Storage". How do I make use of the browser storage? Do you want to added this to the demo.

The view locations are stored in the storage and shared between tabs. The clear storage button allows to come back to the default layout (after a refresh). Does it answer your question?

> * Demo: Put "Settings" and "Keybindings" side by side instead of below each other. Then you see both immediately when you hit "Toggle Panel"

I'm aware the demo is improvable, I try to improve it next time I touch it! :+1: thanks for the suggestion

kaisalmen:
> The view locations are stored in the storage and shared between tabs. The clear storage button allows to come back to the default layout (after a refresh). Does it answer your question?

Yes, can this be observed with browser dev tools?

CGNonofr:
> > The view locations are stored in the storage and shared between tabs. The clear storage button allows to come back to the default layout (after a refresh). Does it answer your question?
> 
> Yes, can this be observed with browser dev tools?

Yes, in Application => Storage => IndexedDB

PR #176, opened 2023-08-30
feat: export services required to add custom keybinds & commands

CompuIves:
This PR exposes some services and constants that would be needed to add new keybindings, actions and menu items imperatively.

### DIFF
--- a/src/monaco.ts
+++ b/src/monaco.ts
@@ -15,6 +15,8 @@ import { DisposableStore, IDisposable, IReference } from 'vs/base/common/lifecyc
 import { Registry } from 'vs/platform/registry/common/platform'
 import { IJSONContributionRegistry, Extensions as JsonExtensions } from 'vs/platform/jsonschemas/common/jsonContributionRegistry'
 import { CommandsRegistry, ICommandService } from 'vs/platform/commands/common/commands'
+import { MenuRegistry, MenuId } from 'vs/platform/actions/common/actions'
+import { KeybindingsRegistry } from 'vs/platform/keybinding/common/keybindingsRegistry'
 import { IJSONSchema } from 'vs/base/common/jsonSchema'
 import { Extensions as ConfigurationExtensions } from 'vs/platform/configuration/common/configurationRegistry'
 import { EditorOptionsUtil } from 'vscode/vs/editor/browser/config/editorConfiguration'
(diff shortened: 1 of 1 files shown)


CGNonofr:
I understand your need. At first, we were using the internal api a lot.

Since then, I tried to limit the usage of the internal api as much as possible to make update easier and stop breaking things at each update

Is there anything you want to do that can't be done using the extension api?

CompuIves:
Throughout our application we want to register new actions to the VSCode API so we can start using its command palette instead of ours. However, we can't know during the initialization of VSCode which commands are keybindings will be added, which is required for an extension (in its manifest). That's why we're looking for an imperative way to add new keybindings. The `CommandsRegistry` is already working for adding new commands imperatively, but we need a `KeybindingsRegistry` to also add the keybindings for it.

CompuIves:
This is interesting, in the demo here: https://github.com/CodinGame/monaco-vscode-api/blob/main/demo/src/main.ts#L122-L132 there is a custom registration for a new action with keybinding. Essentially, we'd need something like this. But it seems like the action is added here through the editor interface?

CGNonofr:
The action register through the editor is only visible from that editor

>  Extensions require keybindings/commands to be defined upfront (in its manifest).

The manifest can easily be generated on the fly though

CompuIves:
> The manifest can easily be generated on the fly though

Is there a way to update the manifest of an existing extension? Or should we reload the extension every time a new command is added?

Our editor currently (un)registers new keybindings or commands conditionally based on conditions that can happen later (like a user signed in). We can scope all of those actions behind context keys, wait for a second after mounting the editor and dynamically generate a manifest, but that would be a significant amount of work for the first version. We have over ~100 actions we'd have to port.

I was hoping initially we could use the same API that VSCode uses internally to register these actions, similar to how VSCode already exposes the CommandsRegistry.

CGNonofr:
Instead of adding and removing action on the fly, you can register them all, and just enable/disable them on demand:

```typescript
const { getApi } = registerExtension({
  name: 'custom-commands',
  publisher: 'me',
  version: '1.0.0',
  engines: {
    vscode: '*'
  },
  contributes: {
    commands: [{
      command: 'customCommand',
      title: 'My custom command'
    }],
    menus: {
      commandPalette: [
        {
          command: 'customCommand',
          when: 'customContextKey'
        }]
    },
    keybindings: [
      {
        command: 'customCommand',
        key: 'ctrl+f1',
        mac: 'cmd+f1',
        when: 'customContextKey'
      }
    ]
  }
}, ExtensionHostKind.LocalProcess)

const api = await getApi()

let enabled = true
setInterval(() => {
  enabled = !enabled
  console.log('enable', enabled)
  void api.commands.executeCommand('setContext', 'customContextKey', enabled)
}, 2000)

api.commands.registerCommand('customCommand', () => {
  console.log('Custom command executed')
})
```

CompuIves:
Right, that's a sensible approach and something that we ultimately want to move to. However, right now the actions in our editor are added on-demand in React components. They register and unregister themselves as components mount and unmount. I realise this is not the way VSCode should work, as it's based on context keys. Also, configuration of keybindings would become weird because you wouldn't be able to configure keybindings for actions while the action is not registered.

However, moving all (50+) commands to this new paradigm will require a major refactor, which makes the adoption of VSCode less incremental and more risky for us.

The way it would work for our v1, is that components can register the keybinding + command using `registerKeybindingAndCommand`, when the component unmounts (or its dependencies change), we call dispose on the `IDisposable` that's returned from it. This way, the command palette from VSCode becomes almost a drop-in replacement for our current solution.

It's understandable if you don't want to put this in the library itself. I can also maintain a fork of this library until we're able to move all custom commands to an upfront specification that works well with our React component system.

CGNonofr (approved):

PR #181, opened 2023-09-01
Simulate an editor group for each standalone editor

### LINKED ISSUE #180: DiagnosticFeature not working after upgrade
Hey again,

We recently upgraded the `monaco-languageclient` package from version `4.0.3` to version `6.3.0-next.2`, & since then the LSP `DiagnosticFeature` stopped working.

After spending some time on trying to understand the source of the issue, i noticed that it was due to some vscode api changes, i'll try to explain my conclusions below:

1. The `vscode/dist/api.js` file has considerably changed & more exactly the `tabs` getter which has been moved to a separate file, below the old implementation:

Screen Capture_select-area_20230831181022

& here is the new one, which is now located at `vscode/dist/vscode/vs/workbench/api/common/extHostEditorTabs.js`:

Screen Capture_select-area_20230831181642

2. As a result to the above change(s), the `tabs` property is now always empty even when having opened monaco editor tabs:

Screen Capture_select-area_20230831183447

3. The diagnostic feature does verify if the current `TextDocument` matches with a visible tab & since the `tabs` list is always empty in our case the check always fails & the diagnostic requests are not sent to the LSP server:

Screen Capture_select-area_20230831183616

I think that providing a way to override the `tabs getter` would fix the issue, still i'm not sure if that's doable on this package side or on the vscode api side, what do you think @kaisalmen ?

CGNonofr:
fix #180

published as @codingame/monaco-vscode-api@1.81.7-next.0

### DIFF
--- a/src/missing-services.ts
+++ b/src/missing-services.ts
@@ -165,6 +165,7 @@ import { ITestResultService } from 'vs/workbench/contrib/testing/common/testResu
 import { IDiagnosticsService, NullDiagnosticsService } from 'vs/platform/diagnostics/common/diagnostics'
 import { INotebookSearchService } from 'vs/workbench/contrib/search/browser/notebookSearch'
 import { ResourceSet } from 'vs/base/common/map'
+import { IEditorGroupView } from 'vs/workbench/browser/parts/editor/editor'
 import { unsupported } from './tools'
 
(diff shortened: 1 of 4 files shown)


kaisalmen (approved):
LGTM

PR #182, opened 2023-09-05
Added more detailed instructions for getting the demo working

EthanP1337:


CGNonofr:
Thank you for contributing to that project!

However, here are some issues with your suggestions:
- The instructions seem window-specific (`\` instead of `/`)
- I'm not sure it's really useful to explain how to clone a git repo or change the current directory
- The demo project is probably missing some dependencies (which are currently used from the root project), feel free to install them
- why using the `--legacy-peer-deps` option?
- Instead of suggesting to change the scripts to add cross-env (which is a nice idea) what about making the change directly?

EthanP1337:
Good Afternoon,
     Yeah those are all great points and I'm going to be re-submitting the
PR with many of those changes in mind. I had approached making the changes
in a bad way originally. I'll be modifying that PR into something more
actually useful soon lol.

Thanks,
Ethan

PR #184, opened 2023-09-07
Node ext host

CGNonofr:
It allows to connect to a remote server to run vscode extensions, plug terminals, bind filesystem and file search and probably more

published as `@codingame/monaco-vscode-api@1.81.8-next.0`


CompuIves:
WOW! This is super super impressive, extremely exciting!!!

kaisalmen (commented):
@CGNonofr Awesome. Did some local testing and it works nicely. 😎 
I want to perform further tests with `monaco-languageclient`, but likely won't have time today. Let's see.

kaisalmen:
@CGNonofr I found an issue.
When I update `monaco-languageclient` to `1.81.8-next.0` the following dependency `windows-process-tree` requires Visual Studio to be present on Windows. `npm i` fails due to this and it even seems not to be used. Can you please remove it? Thank you

kaisalmen:
Even with Visual Studio Community installed and the respective C++ tools the install fails.
If you use `--ignore-scripts`  the install succeeds, but also other scripts won't work (treemending)

kaisalmen:
@CGNonofr One way out of here is to declare `peerDependenciesMeta` and declare some dependency as really optional. Unneeded dependency should be removed and you should think about moving some of the non-optional dependency from `peerDependencies` to `dependencies`.
Like this, it really is not end-user compatible.

CGNonofr:
`windows-process-tree` is used by the server unfortunately... we'll need to implement a multi library publish I guess...

kaisalmen:
> we'll need to implement a multi library publish I guess...

You mean a separate package for the server?

Btw, do you want to upgrade to vscode 1.82.0 first and bring this in later?

CGNonofr:
> Btw, do you want to upgrade to vscode 1.82.0 first and bring this in later?

Oh no, already a new version :D

What do you think about doing it yourself to make sure I'm not the only one able to do it?

kaisalmen:
> What do you think about doing it yourself to make sure I'm not the only one able to do it?

Do you have a guide, some breadcrumbs? 😉

CGNonofr:
> > What do you think about doing it yourself to make sure I'm not the only one able to do it?
> 
> Do you have a guide, some breadcrumbs? 😉

No :fearful: That's an opportunity to write it :D

What I usually do is:
- Get the new vscode commit from the new monaco-editor release version (in the package.json there is a `vscodeRef` field)
- Open the vscode repo, reset to the previous vscodeRef commit
- Apply the current patch: `patch -p1 < /path/to/monaco-vscode-api/scripts/vscode.patch`
- `git stash`
- checkout new vscodeRef commit
- `git stash pop`
- resolve conflicts, update code...
- generate new patch: `git diff > /path/to/monaco-vscode-api/scripts/vscode.patch`

Now on the monaco-vscode-api side:
- update monaco-editor (and other dependencies)
- wait for the new vscode version to be downloaded and built
- Fix errors, adapt code, build
- update demo

That's it I think

kaisalmen:
Nice, that's already quite helpful. Good opportunity to put a new markdown document in ./doc or the root of the repo 🙂 

Btw, what happened to 0.42.0 release of monaco-editor?
image

CGNonofr:
> Nice, that's already quite helpful. Good opportunity to put a new markdown document in ./doc or the root of the repo 🙂
> 
> Btw, what happened to 0.42.0 release of monaco-editor? image

Ahah that's a good question, probably a mistake on their side, they seem to try to release RC versions

kaisalmen (approved):
LGTM. The whole PR stack (this + #191 and #195) is working with `monaco-languageclient`.

PR #185, opened 2023-09-11
Clarify Demo Instructions

EthanP1337:
- I added more detailed instructions for getting the demo working.
- I add a .gitattributes file that sets "* text eol=lf". This prevents git from changing the line endings to the environment the repo was cloned to and enforces the "LF" line ending. This prevents issues with the bash scripts in ./scripts from having the line endings changed.
- I added a "demo troubleshooting" section for troubleshooting the line ending issues mentioned above
- I also cleaned up markdown in the README

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# Declare that all files will always have LF line endings on checkout.
+* text eol=lf
--- a/README.md
+++ b/README.md
@@ -312,7 +312,7 @@ plugins: [
 import './extension.vsix'
 ```
 
-### Demo
+## Demo
 
 Try it out on https://codingame.github.io/monaco-vscode-api/
 
@@ -328,8 +328,15 @@ It includes:
 - Keybinding service, with user keybindings editor
 - Debuggers
 
-From CLI run:
+### Demo Instructions
+First we need to install and build the dependencies the demo will use.
+```bash
+# It will look like nothing is happening, trust the process
+npm install 
+npm run build
+```
 
+Now we can setup and  run the demo
 ```bash
 cd demo
 npm ci
@@ -342,6 +349,20 @@ For the debug feature, also run:
 ```bash
 npm run start:debugServer
 ```
+
+### Troubleshooting Demo Errors
+---
+### Error
+```
+npm ERR! code 127
+npm ERR! path /workspaces/monaco-vscode-api
+npm ERR! command failed
+npm ERR! command sh -c ./scripts/install-vscode && node --loader ts-node/esm src/monaco-treemending.ts && patch-package
+```
+### Fix
+By default, git will change the line endings to whatever environnement the repo was cloned into. If you are getting this error, you need to change the line endings in ./scripts from CRLF to LF. Optionally, you can set ```git config --global core.autocrlf input```
+
+---
 
 ### History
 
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -328,8 +328,15 @@ It includes:
 - Keybinding service, with user keybindings editor
 - Debuggers
 
-From CLI run:
+### Demo Instructions
+First we need to install and build the dependencies the demo will use.
+```bash
+# It will look like nothing is happening, trust the process
+npm install
```
CGNonofr, line 349: `npm ci` instead maybe to not touch the package.lock

README.md
```diff
@@ -328,8 +328,15 @@ It includes:
...
-From CLI run:
+### Demo Instructions
+First we need to install and build the dependencies the demo will use.
+```bash
+# It will look like nothing is happening, trust the process
+npm install 
+npm run build
+```
 
+Now we can setup and  run the demo
```
CGNonofr: (double space)

demo/vite.config.ts
```diff
@@ -2,7 +2,9 @@ import { defineConfig } from 'vite'
 import * as fs from 'fs'
 import url from 'url'
 
-const cdnDomain = 'http://127.0.0.2:5173'
+//const cdnDomain = 'http://127.0.0.2:5173'
+const cdnDomain = 'http://127.0.0.1:5173' // This fixes the issue with CORS
+
```
CGNonofr, line 7: The whole point is to use another domain for workers

README.md
```diff
@@ -329,7 +329,11 @@ It includes:
 - Debuggers
 
 ### Demo Instructions
+
+⚠️ Make sure to open the project in a Dev Container[^1] ⚠️
```
CGNonofr: Is it specific to an environment? I don't need any dev container

.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Declare that all files will always have LF line endings on checkout.
+* text eol=lf
```
CGNonofr, line 2: if the readme section still required then?

demo/vite.config.ts
EthanP1337: Just to run the demo?

CGNonofr: What do you mean?

EthanP1337: It may be something wrong i'm doing on my end but when I run the demo with the cdnDomain set to 'http://127.0.0.2:5173' the icons are unable to load due to CORS

README.md
EthanP1337: Fair enough

.gitattributes
EthanP1337: Good point

demo/vite.config.ts
digitallyserviced: > 'http://127.0.0.2:5173' the icons are unable to load due to CORS

Do you have some non-default security settings or adblocker? 

With the two separate IPs/domains I do not have an issue with CORS and assets, or workers.

image

digitallyserviced:
I think your issue about CORS and the `127.0.0.2` address not resolving/loading assets could be due to the devcontainer?

While all IPs in that class A lead to home, they are not the same address if specifically bound on/to.

I could bind three services to :5173 if I assign one to `127.127.127.127:5173` and the other to `127.254.254.254:5173`, and maybe a third at `127.0.0.1:5173`. They are the same machine, not the same socket.

CGNonofr:
@EthanP1337 What is the status of this PR?

CGNonofr:
ping @EthanP1337

Issue #186, opened 2023-09-12
Failing to load vscode-oniguruma wasm (and other files) since 1.80


jhk-mjolner:
I've tried upgrading from monaco-languageclient 6.0.3 to the latest 6.4.6, and got several errors in the browser console related to not being allowed to load local resources.
I've tried different versions, and concluded that the error first appears in version 6.3.0 which uses 1.80 of monaco-vscode-api.

Error examples from browser console:
`Not allowed to load local resource: file:///C:/<PATH_TO_MY_PROJECT>/node_modules/vscode/dist/service-override/vscode-oniguruma/release/onig.wasm`
`Not allowed to load local resource: file:///C:/<PATH_TO_MY_PROJECT>/frontend/node_modules/vscode/dist/default-extensions/theme-defaults/light_modern.json` (there's several of this kind of error, about trying to load various theme .json files)
And those files rightly aren't there, so webpack can't bundle them and serve them as resources.
The error is thrown from `node_modules\vscode\dist\vscode\vs\workbench\services\textMate\browser\textMateTokenizationFeatureImpl.js` in `async _loadVSCodeOnigurumaWASM()` where it tries to load the path `'vscode-oniguruma/../onig.wasm'`

I've tried looking at some source code and believe this change might be the culprit: https://github.com/CodinGame/monaco-vscode-api/commit/e641ac323eb8e29c9496ab87eae345823a090333 as it was introduced in 1.80, remains in 1.81.7 (latest), and changes how the path to the wasm file is resolved.
Perhaps another similar change in another commit, also introduced in 1.80, might explain the same error for the theme .json file(s)?

I'm using Angular and an `extra-webpack.config.js` file to add webpack configuration and have this part for loading wasm, that worked fine before version 1.80:
```js
      {
        test: /\.(mp3|wasm|ttf)$/i,
        type: 'asset/resource',
      },
      {
        test: /\.wasm$/,
        type: 'asset/inline',
      },
```

Is there some configuration I can use, either for monaco-languageclient, monaco-vscode-api or webpack, that can resolve this issue?
Or does it require code changes from your side?

CGNonofr:
I see you are on windows and I've never tested it on that platform, maybe the `C:/` part breaks it? (at least for the extension files part: `light_modern.json` for instance`)

Regarding the wasm loading issue, did you try to add an alias from `vscode-oniguruma/release/onig.wasm` to the absolute file path in the webpack config?

[...]

PR #188, opened 2023-09-13
Added solution for `Not allowed to load local resource` errors with Angular

jhk-mjolner:
Related to https://github.com/CodinGame/monaco-vscode-api/issues/186

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,20 @@ There are workarounds for both:
 }
 ```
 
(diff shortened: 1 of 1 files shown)


CGNonofr:
That's perfect! Thank you

We'll probably move it in a troubleshooting section in the future when there are more of them

CGNonofr (approved):

PR #194, opened 2023-09-25
fix eslint path in package.json scripts (broke compile script)

digitallyserviced:
So this has bothered me for a few past revs lol. 

I do not know why it hasn't bothered anyone else. It could be because of their OS, eslint version. Or how they are setup and their current $PWD, symlinks, volume bind mount, containers, etc...

But the `npm run lint` has never seemed to work properly with the absolute root path of `/src/**`. 

Since the other paths for eslint specified; `rollup` and `.` , are relative paths, I assumed it would be acceptable for the `src` as well. 

It should still be compatible with other OS's and eslint versions.

However not neccessarily compatible with setups using pecific symlinks, bind mounts, $PWD, bind mounts.

image

It may be somewhat petty, but it's nice when a `npm run compile` can complete and I don't have to press anything else.


CGNonofr:
Thank you :)

CGNonofr (approved):

CGNonofr:
Ok didn't notice the target branch...

PR #206, opened 2023-10-10
Better split service overrides

### LINKED ISSUE #201: main package now requires skipLibCheck=true
Hi @CGNonofr 

because the main because is now only dependent on `monaco-editor` the typescript compilation now only succeeds with `skipLibCheck=true`.

I have reproduced this on two branches with the Angular example repo:

With lib checking enabled:
https://github.com/TypeFox/monaco-languageclient-ng-example/actions/runs/6422702376/job/17439750585

With lib checking disabled:
https://github.com/TypeFox/monaco-languageclient-ng-example/actions/runs/6422712538/job/17439771153

It is the same behaviour with `monaco-languageclient`. With the latest commit on `main` I globally enabled lib checking, but disabled it where it is required (packages client, examples and verify/webpack).

The problematic code/definitions are terminal related things and xterm related dependencies. Do you have a good idea how to fix this?

CGNonofr:
This PR does 3 things:
- It throws an error when trying to initialize the already-initialized service
- It extract the service-override code and types into their own package to make monaco-vscode-api as small as possible
- It makes it so the types of the monaco-vscode-api package don't refer to any external dependency other than monaco-editor or vscode (fix #201)

Can you confirm the issue is fixed @kaisalmen ? published as [1.82.5-next.0](https://www.npmjs.com/package/@codingame/monaco-vscode-api/v/1.82.5-next.0)

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -9,8 +9,8 @@
     "compile": "tsc",
     "start": "NODE_OPTIONS=--experimental-import-meta-resolve vite --config vite.config.ts",
     "start:debug": "vite --config vite.config.ts --debug --force",
-    "build": "vite --config vite.config.ts build",
-    "build:github": "vite --config vite.github-page.config.ts build && touch dist/.nojekyll",
+    "build": "tsc --noEmit && vite --config vite.config.ts build",
+    "build:github": "tsc --noEmit && vite --config vite.github-page.config.ts build && touch dist/.nojekyll",
     "start:debugServer": "node --loader ts-node/esm src/debugServer.ts",
     "start:extHostServer": "vscode-ext-host-server --without-connection-token"
   },
--- a/demo/tsconfig.json
+++ b/demo/tsconfig.json
@@ -16,6 +16,7 @@
     "noImplicitReturns": true,
     "baseUrl": ".",
     "paths": {
+      "vscode/vscode/*": ["../dist/main/vscode/src/*"],
       "vscode/*": ["../dist/main/*"]
     }
   },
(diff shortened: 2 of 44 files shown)


kaisalmen:
@CGNonofr superb! Btw, "generate-types" is out of memory the branch could also be rebased to latest main.
Will tests this with the next version and let you know

kaisalmen:
@CGNonofr first impression is good. But I need to look at it again tomorrow.

kaisalmen:
@CGNonofr I see strange timeout suddenly on the languageclient side. But, I need to first sort whether my latest tsconfig "clean-up" are responsible for this.

kaisalmen (dismissed):
@CGNonofr changes work as expected in `monaco-languageclient` 👍 

One thing about the rollup scripts: Does it make sense to extract common functions from the `rollup.config.ts` and `rollup.types.config.ts`? Or is it not worth the effort. This is not blocking. Hust a thought.

CGNonofr:
Build should be fixed

> @CGNonofr changes work as expected in `monaco-languageclient` 👍

Awesome!

> One thing about the rollup scripts: Does it make sense to extract common functions from the `rollup.config.ts` and `rollup.types.config.ts`? Or is it not worth the effort. This is not blocking. Hust a thought.

I just had a look and I don't really know how to factorize, it seems duplicated but there is a lot of small tweaks as the whole rolllup-plugin-dts is a hack

btw: 4 new commits

kaisalmen:
> btw: 4 new commits
> 
> published as 1.82.5-next.1

Will test this now. I do only test a fraction of the produced npm modules. Any idea how to verify everything is working (maybe automatically?) The demo imports everything, right?

kaisalmen (approved):
LGTM

PR #208, opened 2023-10-12
Check already patched before checking version

CGNonofr:


kaisalmen (approved):
LGTM

CGNonofr:
@kaisalmen By doing so, updating monaco-vscode-api without updading monaco-editor will no longer throw an error, isn't that an issue?

kaisalmen:
The `peerDependency` must be specific. E.g. next version (`1.83.0` must have `"monaco-editor": "0.44.0"` as `peerDependency` than every dependent package should get a warning, right?

CGNonofr:
> The `peerDependency` must be specific. E.g. next version (`1.83.0` must have `"monaco-editor": "~0.44.0"` as `peerDependency` than every dependent package should get a warning, right?

We don't use peerDependencies anymore but I got your point!

Most users ignore those errors though and we introduced the version check to prevent this

kaisalmen:
> Most users ignore those errors though and we introduced the version check to prevent this

Bold idea: `monaco-vscode-api` only relies on `@codingame/monaco-editor-treemended` and people should use it directly. Bundlers should use WA like this [one we suggested for monaco-editor-core](https://github.com/TypeFox/monaco-languageclient/tree/main#monaco-editor-core) in the past.

Maybe, this clearer. There is no ideal solution, but it would move the problem to become just a configuration issue

kaisalmen:
@CGNonofr what about a check a runtime that directly logs at a very early point an info or an error if a tree-mended monaco-editor or not is used. This should be fairly easy to add, right?

CGNonofr:
> @CGNonofr what about a check a runtime that directly logs at a very early point an info or an error if a tree-mended monaco-editor or not is used. This should be fairly easy to add, right?

I'm not sure, how would you implement it?

kaisalmen:
> I'm not sure, how would you implement it?

Augment `MonacoEnvironment` in the treemended package and check it at initialize?

CGNonofr:
Why not, it may help! feel free to implement it :+1:

kaisalmen:
> Why not, it may help! feel free to implement it 👍

Ok, but it won't happen today

PR #220, opened 2023-10-24
Remove non-used traductions

CGNonofr:
Only keep traductions from 586 modules over 995

Traductions files go (uncompressed) from ~1.1mo to ~700ko

### DIFF
--- a/rollup/rollup.language-packs.ts
+++ b/rollup/rollup.language-packs.ts
@@ -3,6 +3,7 @@ import * as rollup from 'rollup'
 import { IExtensionManifest } from 'vs/platform/extensions/common/extensions'
 import { dataToEsm } from '@rollup/pluginutils'
 import { PackageJson } from 'type-fest'
+import glob from 'fast-glob'
 import * as fs from 'fs'
 import * as path from 'path'
 import { fileURLToPath } from 'url'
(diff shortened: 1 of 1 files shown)


CGNonofr:
> Sounds great and will result in smaller app bundles in the end.

It's still no as compact as in VSCode, they have a VERY complex build process that transforms the parameter of every calls to `localize` by a unique autoincremented id. I don't think we can do the same easily but maybe something based on hashes... I don't know...

kaisalmen:
@CGNonofr will be interesting to see how much the reduction already is for rollup bundles.

kaisalmen:
@CGNonofr The bundle size in the `monaco-languageclient` vite verification went from 6865 kb down to 6850 kb. That's 15 kb. The effect on application level is not as big as expected.

CGNonofr:
How is that possible?

kaisalmen:
> How is that possible?

Good question. Does rollup remove unneeded code effectively?

CGNonofr:
> > How is that possible?
> 
> Good question. Does rollup remove unneeded code effectively?

That isn't even rollup but my code that does it, and I measure much more than that

kaisalmen:
Yes, I meant `vite build` on the previous version (1.83.3) seemed to be already effective then, right?

CGNonofr:
if you compare https://registry.npmjs.org/@codingame/monaco-vscode-language-pack-fr/-/monaco-vscode-language-pack-fr-1.83.4.tgz with https://registry.npmjs.org/@codingame/monaco-vscode-language-pack-fr/-/monaco-vscode-language-pack-fr-1.83.3.tgz, it's 253ko vs 337ko compressed

I don't see how any treeshaking algorithm would be able to improve it without help

kaisalmen:
> I don't see how any treeshaking algorithm would be able to improve it without help

Me neither. Those dependencies go into the bundle:
image

That is the output:
image

kaisalmen:
@CGNonofr not all of the above deps really end up in the bundle. E.g. the language packs aren't used in the json client. Where there substantial reduction for the main and monaco-editor packages?

CGNonofr:
I don't think anything else changed

PR #227, opened 2023-11-07
fix: swap toRemove & toAdd in deltaExtensions

CompuIves:
When we try to remove & add the same extension in a very short timeframe, it does not get initialized properly because it gets both added to `toRemove` and `toAdd`. When that happens, the extension is first added, and then immediately removed. By swapping `toRemove` and `toAdd` we ensure that the extension is first removed, and then newly added.

Here's similar code in VSCode where I think they encountered the same condition:

image


CGNonofr:
Thank you!

CGNonofr (approved):

CGNonofr:
Let wait until all PRs are merged before releasing a new version

PR #228, opened 2023-11-07
feat: add support for custom fs provider

CompuIves:
We're looking to add our own providers for `user` and `cache` fs so they can be persisted between sessions. With this change, we don't break the existing API but we can allow for adding custom filesystems for these providers.

### DIFF
--- a/src/service-override/files.ts
+++ b/src/service-override/files.ts
@@ -13,6 +13,8 @@ import { HTMLFileSystemProvider } from 'vs/platform/files/browser/htmlFileSystem
 import * as path from 'vs/base/common/path'
 import 'vs/workbench/contrib/files/browser/files.contribution.js?include=registerConfiguration'
 import { Schemas } from 'vs/base/common/network'
+import { IndexedDBFileSystemProvider } from 'vs/platform/files/browser/indexedDBFileSystemProvider'
+import { IndexedDB } from 'vs/base/browser/indexedDB'
 
 abstract class RegisteredFile {
   private ctime: number
@@ -534,6 +536,8 @@ export {
   FilePermission,
   HTMLFileSystemProvider,
   InMemoryFileSystemProvider,
+  IndexedDB,
+  IndexedDBFileSystemProvider,
   RegisteredFileSystemProvider,
   RegisteredFile,
   RegisteredReadOnlyFile,
--- a/src/service-override/files.ts
+++ b/src/service-override/files.ts
@@ -458,23 +458,30 @@ const fileSystemProvider = new OverlayFileSystemProvider()
 fileSystemProvider.register(0, new MkdirpOnWriteInMemoryFileSystemProvider())
 
 const extensionFileSystemProvider = new RegisteredFileSystemProvider(true)
+
+const providers: Record<string, IFileSystemProvider> = {
+  user: new InMemoryFileSystemProvider(),
+  extension: extensionFileSystemProvider,
+  cache: new InMemoryFileSystemProvider(),
+  logs: new InMemoryFileSystemProvider(),
+  [Schemas.vscodeUserData]: new InMemoryFileSystemProvider(),
+  [Schemas.tmp]: new InMemoryFileSystemProvider(),
+  file: fileSystemProvider
+}
+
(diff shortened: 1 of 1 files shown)


src/service-override/files.ts
```diff
@@ -457,30 +457,51 @@ class DelegateFileSystemProvider implements IFileSystemProvider {
...
+      let fileSystemProviderDisposable = this.registerProvider('file', fileSystemProvider)
+      fileSystemProvider.onDidChangeOverlays(() => {
+        fileSystemProviderDisposable.dispose()
+        fileSystemProviderDisposable = this.registerProvider('file', fileSystemProvider)
+      })
+    }
+
+    for (const [scheme, provider] of Object.entries(additionalProviders)) {
+      this.registerProvider(scheme, provider)
+    }
```
CGNonofr, line 498: What about making it a little more compact (and not repeat the schema each time) with something like that (not tested)?
```typescript
const providers = {
  user: new InMemoryFileSystemProvider(),
  extension: extensionFileSystemProvider,
  cache: new InMemoryFileSystemProvider(),
  logs: new InMemoryFileSystemProvider(),
  [Schemas.vscodeUserData]: new InMemoryFileSystemProvider(),
  [Schemas.tmp]: new InMemoryFileSystemProvider(),
  ...additionalProviders
}
for (const [scheme, provider] of Object.entries(providers)) {
  let disposable = this.registerProvider(scheme, provider)
  if (provider instanceof OverlayFileSystemProvider) {
    provider.onDidChangeOverlays(() => {
      disposable.dispose()
      disposable = this.registerProvider('file', fileSystemProvider)
    })
  }
}

```

CompuIves: I like it! Will update the PR.

CompuIves: Updated!

CompuIves:
I'm also exposing `IndexedDBFileSystemProvider`, which I want to use for the user store. Is it okay if that's part of this PR?

CompuIves:
While I'm testing the code, I can't seem to get the `customProviders` inside the constructor. I was under the assumption that if it's passed to the `staticArguments` array, it would be passed as the second argument in the constructor, but it's empty for me. Have you seen something like this before?

CompuIves:
Ah, found that static arguments need to be passed as first argument, and then the decorators. I'll try that now.

CompuIves:
That didn't work either. It looks like another place calls `createInstance` on it (in my case, the simple extension service) and that causes the static argument to be lost. I have another way we can solve this, but it's less nice. I'll commit that as part of the first commit now.

CompuIves:
Ah, I found the issue. The `getServiceOverride` is called multiple times, also by other services. Because of this, I've decided to create a specific function to override the different filesystem providers. That ensures that we don't accidentally reset them.

src/service-override/files.ts
```diff
@@ -484,6 +498,15 @@ export default function getServiceOverride (): IEditorOverrideServices {
...
 }
 
+/**
+ * Register a custom file system provider for the given schemes (the keys are the schemes,
+ * like "http", or "file"). This function needs to be called _before_ the file service is
+ * instantiated.
+ */
+export function registerCustomProviders (providers: CustomFileSystemProviders): void {
+  customFileSystemProviders = providers
+}
```
CGNonofr, line 510: What about being consistent with the `registerFileSystemOverlay` function by transforming it into `registerCustomProvider(scheme: string, provider)`

It could also simplify a bit the code:
```typescript
const providers = {
  user: new InMemoryFileSystemProvider(),
  extension: extensionFileSystemProvider,
  cache: new InMemoryFileSystemProvider(),
  logs: new InMemoryFileSystemProvider(),
  [Schemas.vscodeUserData]: new InMemoryFileSystemProvider(),
  [Schemas.tmp]: new InMemoryFileSystemProvider(),
  file: fileSystemProvider
}
export function registerCustomProvider (scheme: string, provider: transforming): void {
  customFileSystemProviders[scheme] = provider
}
```

CompuIves: Good suggestion! I made the suggested changes.

CGNonofr (approved):

PR #229, opened 2023-11-07
feat: add ability for creating custom editors for files

CompuIves:
This PR is a bit more tricky. I've been looking into ways that we can render custom editors for files. Our specific use case is rendering a custom interactive Markdown for markdown files, and we've been wanting to use React for this as it's more integrated with our editor.

While there's currently a nice API for creating custom editors, it does not handle creating custom editors for files yet. The missing piece to enable this is the `IEditorResolverService`, you can register a custom editor like so:

```ts
  const registerEditor = ({
    id,
    name,
    onRender,
    globPattern,
  }: RegisterEditorParams) => {
    const disposableStore = new DisposableStore();

    // First create our custom editor
    const { disposable, CustomEditorInput } = registerEditorPane({
      id: id,
      name,
      renderBody: (container) => {
        return onRender(container);
      },
    });
    disposableStore.add(disposable);

    // Then register that editor for the editor resolver service, to tell VSCode to use this
    // editor when the globPattern matches (e.g. `*.md`, or `csb-file:/**`).
    const editorResolverService = StandaloneServices.get(
      IEditorResolverService
    );
    const resolveDisposable = editorResolverService.registerEditor(
      globPattern,
      {
        id,
        label: name,
        // @ts-ignore
        priority: "default" as const,
      },
      {},
      {
        createEditorInput: (baseInput) => {
          return {
            options: {},
            editor: new CustomEditorInput(undefined, baseInput, id),
          };
        },
      }
    );
    disposableStore.add(resolveDisposable);

    return disposableStore;
  };
```

This works really well, but the last missing piece is that VSCode will call `setInput` on the editor to set the actual input. It does this, because it reuses editors and swaps the underlying model/input. We need a way to respond to that `setInput` call, and so I extended the `onRender` call to not just respond with a disposable, but also with an optional `setInput`.


src/service-override/views.ts
```diff
@@ -241,7 +256,7 @@ function registerEditorPane (options: EditorPanelOption): { disposable: IDisposa
     private description: Label = options.name
     private dirty: boolean = false
 
-    constructor (public override readonly closeHandler?: IEditorCloseHandler) {
+    constructor (public override readonly closeHandler?: IEditorCloseHandler, public baseInput?: IResourceEditorInput | ITextResourceEditorInput | undefined, private internalEditorId?: string | undefined) {
```
CompuIves, line 259: `baseInput` is needed for the `resource` call which VSCode uses to deduplicate editors based on resources (I think). This is the description:

image

CGNonofr:
Can you please fix the eslint warning and the demo?

CGNonofr:
a little bit complex indeed :)

What about adding the `globPattern` parameter to the `registerEditorPane` function? Do you think it makes sense?

Also can you demonstrate it in the demo?

CompuIves:
> Do you think it makes sense?

The user would lose some customisability (e.g. can't pass editor options, or define how the untitled editor should launch). But at the same time, it would make the API more accessible. I can add it as an optional parameter. If people then don't pass it they still have the opportunity to register it in a custom way themselves 😄.

CGNonofr:
> > Do you think it makes sense?
> 
> The user would lose some customisability (e.g. can't pass editor options, or define how the untitled editor should launch). But at the same time, it would make the API more accessible. I can add it as an optional parameter. If people then don't pass it they still have the opportunity to register it in a custom way themselves 😄.

Yep, it looks like a good compromise

CompuIves:
How can I get access to the `editorResolverService` from within the function? Is there a way that I can get access to the instantiationService?

CGNonofr:
> How can I get access to the `editorResolverService` from within the function? Is there a way that I can get access to the instantiationService?

StandaloneServices.get is fine. You just need to make sure the services are initialized, either by wrapping it into a `StandaloneServices.withServices` or using `getService` from `services.ts`

CompuIves:
Okay, will apply the changes in about ~45min!

CGNonofr:
> Okay, will apply the changes in about ~45min!

TIC TAC TIC TAC :smile:

CompuIves:
On it! 😂

CompuIves:
Okay, updated, but I can't get the demo running, so will need to double check that.

CGNonofr:
> Okay, updated, but I can't get the demo running, so will need to double check that.

What is the error?

CompuIves:
This one: 

image

When building

CompuIves:
Okay, this change is not working yet. Because Monaco triggers `onDidInitialize` before the storage service is initialized, some services get initialized.

CompuIves:
Or... No, our `withServices` causes the initialization to happen to soon. So as long as we don't use `withServices`, we can make it work.

CGNonofr:
Not sure to follow, what is the solution?

maybe you can use `getService` which also wait for service initialization participants

CompuIves:
Ah, `getService` sounds like the right solution. I'll try that. The problem right now is that with `withServices` my callback gets executed before the participants have been run, and when I instantiate the `editorResolverService` it cannot access the `storageService` yet.

CompuIves:
Okay, confirmed that with this push everything is working!

demo/vite.config.ts
```diff
@@ -3,7 +3,7 @@ import * as fs from 'fs'
 import url from 'url'
 import path from 'path'
 
-const cdnDomain = 'http://127.0.0.2:5173'
+const cdnDomain = 'http://127.0.0.1:5173'
```
CompuIves, line 6: @CGNonofr Are you okay with this change? This was required on my side to make the icons/extensions work. I can imagine that maybe you have `127.0.0.1` as localhost?

CompuIves: I can change it back easily if we want to merge this PR first.

CGNonofr:
I'm waiting for this PR before creating a new release, can we work on merging it asap? :)

CompuIves:
Oh it's ready to be merged! Or is there something else?

CGNonofr:
There is still a few comments from me you didn't address :)

CompuIves:
Ah, hmm. So for these comments:

> What about adding the globPattern parameter to the registerEditorPane function? Do you think it makes sense?

The `registerEditorPane` now has a `globPattern` argument, but it's behind a specific options object. Allows us to expand it with more customization if we want to in the future.

> Also can you demonstrate it in the demo?

The existing custom editor is now also opened for `.md` files.

Are there some other comments that I missed?

CGNonofr:
Yes there is

image

CompuIves:
Oh they're still on "Pending"! So I can't see them yet, I'll work on them from the screenshot.

CompuIves:
Addressed the feedback!

CGNonofr:
> Oh they're still on "Pending"! So I can't see them yet, I'll work on them from the screenshot.

Ok totally my bad... sorry, I'm more used to gitlab than github

src/service-override/views.ts
```diff
@@ -304,16 +337,41 @@ function registerEditorPane (options: EditorPanelOption): { disposable: IDisposa
...
       options.id,
       options.name
     ),
-    [new SyncDescriptor(CustomEditorInput)])
+    [new SyncDescriptor(CustomEditorInput)]))
+
+  const registerEditorOptions = options.registerEditorOptions
+  if (registerEditorOptions != null) {
+    getService(IEditorResolverService).then(editorResolverService => {
+      disposableStore.add(editorResolverService.registerEditor(
```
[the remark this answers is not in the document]
CompuIves: Oh! I always thought that `DisposableStore` would automatically dispose new additions after it was disposed, but that's not the case after double checking:

image

Misremembered. I can do an early return here.

demo/src/features/customView.ts
```diff
@@ -54,8 +54,19 @@ const { CustomEditorInput } = registerEditorPane({
...
+        if (input.resource != null) {
+          container.innerHTML = 'Opened file: ' + input.resource.path
+        } else {
+          container.innerHTML = 'This is a custom editor pane<br />You can render anything you want here'
+        }
       }
     }
+  },
+  registerEditorOptions: {
+    globPattern: '*.md'
```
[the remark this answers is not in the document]
CompuIves: I can make it work with other files as well! Will change it.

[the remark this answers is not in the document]
CompuIves: Updated!

src/service-override/views.ts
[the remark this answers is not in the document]
CompuIves: Added early return!

src/service-override/views.ts
```diff
@@ -210,6 +230,7 @@ interface SimpleEditorInput extends EditorInput {
 function registerEditorPane (options: EditorPanelOption): { disposable: IDisposable, CustomEditorInput: new (closeHandler?: IEditorCloseHandler) => SimpleEditorInput } {
```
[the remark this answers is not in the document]
CompuIves: Parameter is added

src/service-override/views.ts
```diff
@@ -250,6 +265,9 @@ function registerEditorPane (options: EditorPanelOption): { disposable: IDisposa
     }
 
     override get resource (): URI | undefined {
+      if (this.baseInput != null) {
+        return this.baseInput.resource
+      }
```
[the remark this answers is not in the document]
CompuIves: Done

CompuIves:
Okay, addressed the remaining feedback

src/service-override/views.ts
```diff
@@ -207,9 +227,10 @@ interface SimpleEditorInput extends EditorInput {
   setDirty (dirty: boolean): void
 }
 
-function registerEditorPane (options: EditorPanelOption): { disposable: IDisposable, CustomEditorInput: new (closeHandler?: IEditorCloseHandler) => SimpleEditorInput } {
+function registerEditorPane (options: EditorPanelOption): { disposable: IDisposable, CustomEditorInput: new (closeHandler?: IEditorCloseHandler, baseInput?: IResourceEditorInput | ITextResourceEditorInput) => SimpleEditorInput } {
```
CGNonofr, line 230: doesn't `ITextResourceEditorInput` extends `IResourceEditorInput`?

src/service-override/views.ts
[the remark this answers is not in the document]
CGNonofr: perfect, I didn't think of that solution

src/service-override/views.ts
CompuIves: Right, I've taken this from the editor input factory that needs to be passed to the editor resolver:

image

Which is used here and passed as `baseInput`:

image

I can change it to only use `IResourceEditorInput`, but that would make it a bit less consistent with the API of VSCode.

CompuIves: I do agree that we can just accept `IResourceEditorInput`. The only reason to keep the union is to indicate that this is the type that VSCode passes to the `createEditorInput`.

src/service-override/views.ts
```diff
@@ -197,7 +203,21 @@ type Label = string | {
 interface EditorPanelOption {
   readonly id: string
   name: string
-  renderBody (container: HTMLElement): IDisposable
+  renderBody (container: HTMLElement): BodyRenderer
```
CGNonofr, line 206: wouldn't it be more simple to have the `setInput` method next to the `renderBody` method?

CGNonofr: is it possible that `renderBody` is called without `setInput`? Can't we just provide the input as a parameter of `renderBody` instead of having a setInput method?

src/service-override/views.ts
```diff
@@ -189,6 +191,10 @@ function renderStatusBarPart (container: HTMLElement): IDisposable {
   return attachPart(Parts.STATUSBAR_PART, container)
 }
 
+interface BodyRenderer extends IDisposable {
+  setInput? (input: EditorInput, options: IEditorOptions | undefined, context: IEditorOpenContext, token: CancellationToken): Promise<void>
```
CGNonofr, line 195: What about providing a service accessor as parameter? otherwise there is no simple way to get the FileService to get the file content

src/service-override/views.ts
```diff
@@ -304,16 +334,45 @@ function registerEditorPane (options: EditorPanelOption): { disposable: IDisposa
...
+        {
+          createEditorInput (editorInput: IResourceEditorInput | ITextResourceEditorInput, _group: IEditorGroup) {
+            return {
+              options: {},
+              editor: new CustomEditorInput(undefined, editorInput)
+            }
+          }
+        }
+      ))
+    }).catch(console.error)
```
CGNonofr, line 371: I'm not really happy with having an async method the caller is not able to await nor catch errors

Instead, what about returning an async `registerEditor(globPattern: string, options?: Partial<Omit<RegisteredEditorInfo, 'id'>>)` method, next to the `disposable` and `CustomEditorInput`? That would also allow to register multiple glob patterns for a single editorPane

What do you think?

src/service-override/views.ts
```diff
@@ -304,16 +334,45 @@ function registerEditorPane (options: EditorPanelOption): { disposable: IDisposa
...
+          id: CustomEditorInput.ID,
+          label: options.name,
+          priority: registerEditorOptions.priority ?? RegisteredEditorPriority.default
+        },
+        {},
+        {
+          createEditorInput (editorInput: IResourceEditorInput | ITextResourceEditorInput, _group: IEditorGroup) {
+            return {
+              options: {},
+              editor: new CustomEditorInput(undefined, editorInput)
```
CGNonofr, line 366: That's a shame we can't provide a `IEditorCloseHandler` in that case :)

CompuIves:
I like these suggestions. I have a call right now, will go through them in 30min and apply them one by one.

src/service-override/views.ts
CompuIves: > is it possible that renderBody is called without setInput? Can't we just provide the input as a parameter of renderBody instead of having a setInput method?

Unfortunately this is not possible. VSCode first creates the editor, and a couple ticks later it will call `setInput`. So it's not synchronously available. This was my first attempt, but after this I realised that the same editor will be reused between inputs.

CGNonofr: What about making the input parameter optional? (at the first render, it's undefined, then it's called every time the input change with the new input)

CompuIves: Interesting... So I think for us that would be less interesting, because we render a React component for the editor, and then when the input changes we don't rerender the whole editor, but we change the prop of the component itself.

Hmm, though we _can_ check if it's already rendered, and then update the props if it's there. I'll take a look.

CGNonofr: I'm interested on how you implemented it with React, that was my concern, how do you update your existing component prop in the callback?

CompuIves: I'm using portals, which allows me to render React components in the given element by `onRender`, while still in the same React context.

Code will say a thousand words:

```tsx
export const CustomEditor = ({
  globPattern,
  id,
  name,
  Component,
  getName,
}: CustomEditorProps) => {
  const [containers, setContainers] = React.useState<Container[]>([]);
  const { monacoEditor } = useEnvironmentInterface();
  const lastId = React.useRef(0);

  React.useEffect(() => {
    const disposablePromise = monacoEditor.registerEditor({
      id,
      name,
      globPattern,
      onRender: (element) => {
        const id = lastId.current++;
        setContainers((containers) => [
          ...containers,
          {
            id,
            element,
          },
        ]);

        return {
          dispose: () => {},
          setInput: async (editorInput) => {
            if (
              "setName" in editorInput &&
              typeof editorInput.setName === "function" &&
              getName
            ) {
              // This is a hacky way to set the title to the filename, but it's the
              // best way right now to set it. Ideally we use `EditorResourceAccessor`.

              editorInput.setName(getName(editorInput.resource));
            }

            setContainers((containers) =>
              containers.map((container) =>
                container.id === id
                  ? { ...container, filePath: editorInput.resource }
                  : container
              )
            );
          },
        };
      },
    });

    return () => {
      disposablePromise.then((disposable) => disposable.dispose());
    };
  }, []);

  return (
    <>
      {containers.map((container) =>
        createPortal(
          <Box
            css={{
              display: "block",
              width: "100%",
              height: "100%",
              padding: "$2",

              boxSizing: "border-box !important",
              "*": {
                boxSizing: "border-box !important",
              },
            }}
          >
            <Component key={container.id} resource={container.filePath} />
          </Box>,
          container.element
        )
      )}
    </>
  );
};
```

CompuIves: Actually this reminds me that I should still implement dispose.

CompuIves: That renders this:

image

How I use the API:

```tsx
      <CustomEditor
        Component={({ resource }) => {
          if (!resource) {
            return <div>Loading...</div>;
          }

          const relativePath =
            pitcher.clients.fs.absoluteToRelativeWorkspacePath(
              resource.path as AbsoluteWorkspacePath
            );

          return (
            <div style={{ width: "100%", height: "100%" }}>
              <FileMarkdownPreview filepath={relativePath} />
            </div>
          );
        }}
        getName={(resource) => {
          if (resource) {
            return resource.path.split("/").pop() ?? "";
          }

          return "Markdown";
        }}
        globPattern="*.md"
        id="codesandbox.editors.markdown"
        name="Interactive Markdown"
      />
```

CGNonofr: Thank you, interesting :)

src/service-override/views.ts
CompuIves: Actually... Thinking about it, we can enable that, but then it has to be passed to `registerEditorPane`. I can make that change if you want.

CGNonofr: or passed to the potentially new `registerEditor` returned function (:

CompuIves:
Made some changes to the API! This API is _much_ better, I like the suggestion.

src/service-override/views.ts
CompuIves: Will add thise!

CompuIves: Done!

src/service-override/views.ts
```diff
@@ -197,7 +203,7 @@ type Label = string | {
 interface EditorPanelOption {
   readonly id: string
   name: string
-  renderBody (container: HTMLElement): IDisposable
+  renderBody (container: HTMLElement): BodyRenderer
```
CGNonofr, line 206: I'm not really happy with extending IDisposable here

an alternative that is used a lot inside VSCode itself is to provide a DisposableStore as the last parameter

src/service-override/views.ts
```diff
@@ -304,17 +340,39 @@ function registerEditorPane (options: EditorPanelOption): { disposable: IDisposa
...
     ),
-    [new SyncDescriptor(CustomEditorInput)])
+    [new SyncDescriptor(CustomEditorInput)]))
 
   return {
-    disposable,
-    CustomEditorInput
+    disposable: disposableStore,
+    CustomEditorInput,
+    async registerEditor (globPattern: string, registerEditorOptions: RegisterEditorOptions = {}): Promise<IDisposable> {
```
CGNonofr, line 356: At the end, that's a shame to return a Promise here because it's harder to handle, you need to store the promise and await/handle error when you want to dispose it.

I don't see any reason to call that api before the services are initialized anyway, what about checking that the services are initialized (throw an error otherwise?) instead and use `StandaloneServices.get`

CGNonofr (commented):
I'm sorry to have so many concerns, but that MR is a little tricky and it impacts the public API of the library. I'd like it to be as nice as possible because it's better to not change it afterward

src/service-override/views.ts
CompuIves: The risk I see with that is that it's easy to forget for the consumer. If they forget to handle that parameter, then it won't get disposed. Now the type system will give an error if `dispose()` is not implemented on the returned value, making sure that things get cleaned up when we get rid of them.

CGNonofr: The dispose function being mandatory in on the `cons` column for me :smile: most of the time, you don't need it (for instance if you render a react component using portal?)

src/service-override/views.ts
```diff
@@ -304,17 +340,39 @@ function registerEditorPane (options: EditorPanelOption): { disposable: IDisposa
...
+      const resolverService = await getService(IEditorResolverService)
+      return resolverService.registerEditor(
```
CGNonofr, line 358: shouldn't the returned disposable be disposed when the whole editor pane is disposed?

CompuIves:
Will put the changes in today!

src/service-override/views.ts
CompuIves: In our React case we do need to handle dispose, because when dispose is called we need to remove the component from our list. Otherwise the component will try to render on an element that is not attached to the DOM.

CompuIves:
Addressed the feedback and pushed it. Only thing I haven't addressed is the disposable. I would still prefer that disposing is opt-out, instead of opt-in. Otherwise we create a footgun where users can have lingering resources that never get cleaned up even though the editor gets disposed.

src/service-override/views.ts
CompuIves: As an example, in our React case I actually realised that I forgot to properly dispose when I shared the code with you, because the dispose body was empty. I've since changed the code to this:

image

CompuIves: As a last point, I think that this can feel like VSCode, with how they often use classes for UI. E.g. you could also pass this to it:

```ts
class MarkdownRenderer extends Disposable {
  constructor() { }
  
  setInput(input) {
    this.draw();
  }
  
  private draw() { /* ... */ }
}
```

CompuIves:
Quick check on the PR, and I love it. More power to the implementor. I will check more thoroughly in 30min.

CompuIves:
Okay, I like the PR! It's a good approach, and from what I can see so far, it will also work for our case. I particularly like that the tab title can stay consistent.

> (even though it's probably already an already advanced need).

Agree, the closer to VSCode the better. People can always build a simple wrapper around it (probably we'll do this to make it work with React properly)

> it requires experimentalDecorators to be enabled

Will check this, I think that should also be fine on our side!

CompuIves:
Yep, works here!

PR #230, opened 2023-11-08
fix: use same paths as vscode for settings

CompuIves:
VSCode stores its settings in these places (with schema `user-store`):

image

This PR changes our settings to store them there as well, so custom file providers for user data store can put/retrieve settings from there. This should make it possible to persist settings between sessions, if users decide to replace the `user-store` schema with e.g. an IndexedDB filesystem.

It would've been even better if we would get the `userRoamingDataHome` variable directly from the `EnvironmentService`, but since that service is only initialized lazily this is the best next thing.

Code is inspired from here: https://github.com/microsoft/vscode/blob/8cb6def308c41b2b2fd09c14fbbc49d3ec3d5d01/src/vs/platform/userDataProfile/common/userDataProfile.ts#L286-L289

### DIFF
--- a/demo/src/main.ts
+++ b/demo/src/main.ts
@@ -95,7 +95,7 @@ diagnostics.set(modelRef.object.textEditorModel!.uri, [{
   code: 42
 }])
 
-const settingsModelReference = await createModelReference(monaco.Uri.from({ scheme: 'user', path: '/settings.json' }), `{
+const settingsModelReference = await createModelReference(monaco.Uri.from({ scheme: 'user-store', path: '/User/settings.json' }), `{
   "workbench.colorTheme": "Default Dark+",
   "workbench.iconTheme": "vs-seti",
   "editor.autoClosingBrackets": "languageDefined",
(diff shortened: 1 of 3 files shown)


CGNonofr:
Thank you!

is the `user` filesystem provider still required then?

CompuIves:
Good point, I've removed those filesystem providers and also updated the references in the demo to use the new location.

CGNonofr:
> It would've been even better if we would get the userRoamingDataHome variable directly from the EnvironmentService, but since that service is only initialized lazily this is the best next thing.

I think there is a way:
- create the profile in InjectedUserDataProfileService construtor
- inject `IUserDataProfileService` in `UserDataProfilesService`
- use `userDataProfilesService.currentProfile()`

CGNonofr:
Can you please rebase/resolve conflicts? :)

CompuIves:
> I think there is a way

Yep, that did the trick. I've updated them now.

CGNonofr:
Why was the cache fs provider removed?

CGNonofr (approved):

Issue #231, opened 2023-11-08
BrowserStorageService is not initialized


CompuIves:
When we initialize the storage service (using the `StorageService` override), it initialized the `BrowserStorageService` through the `InjectedBrowserStorageService`. However, right now that's not enough to initialize the actual service. To initialize it, we also need to call `initialize()` on the service, like they do here in VSCode code:

```ts
	private async createStorageService(workspace: IAnyWorkspaceIdentifier, logService: ILogService, userDataProfileService: IUserDataProfileService): Promise<IStorageService> {
		const storageService = new BrowserStorageService(workspace, userDataProfileService, logService);

		try {
			await storageService.initialize();

			// Register to close on shutdown
			this.onWillShutdownDisposables.add(toDisposable(() => storageService.close()));

			return storageService;
		} catch (error) {
			onUnexpectedError(error);
			logService.error(error);

			return storageService;
		}
	}
```

Currently we don't do this, and because of this the cache data of e.g. the theme is not cached. I would like to fix this, but I'm not 100% sure how to approach this. The `BrowserStorageService` is initialized in a constructor, which is not async. I wanted to open this issue to discuss approaches.

CGNonofr:
There is a type issue, but it's supposed to be done here => https://github.com/CodinGame/monaco-vscode-api/blob/3b815764fcbd630945c995e49c4e08044217f652/src/service-override/storage.ts#L134

What do you think is the issue with it?

CompuIves:
Very interesting... So from my debugging, it looks like the storage service is accessed already before that participant is executed. The storage service gets called during `initializeMonacoService` by the `WorkbenchThemeService` in its constructor:

<img width="665" alt="image" src="https://github.com/CodinGame/monaco-vscode-api/assets/587016/4a311002-54f6-4d77-baa6-18850f7bc38e">

`WorkbenchThemeService` gets instantiated because of the editor features that are created, looks like SemanticsToken feature needs it:

<img width="645" alt="image" src="https://github.com/CodinGame/monaco-vscode-api/assets/587016/7151a512-aaf8-4d09-9e32-07f686ea91de">

This in turn causes VSCode to default to white theme for a moment. After the constructor of `WorkbenchThemeService` is called, the pre participant is called with the initialisation.

Tricky issue... In VSCode the `BrowserStorageService` is initialised "manually" without dependency injection, and then added to the services. In this case that's a bit harder I reckon.

[...]

PR #233, opened 2023-11-09
fix: initialize editor features after participants have been run

CompuIves:
Fixes https://github.com/CodinGame/monaco-vscode-api/issues/231

With this change, we ensure that editor features are only initialized _after_ we've run our init participants. This ensures that the `WorkbenchThemeService` is not initialized before the `StorageService` is initialized, because `WorkbenchThemeService` already calls to the `StorageService` in its constructor.

The risky thing here is that we're changing how StandaloneServices works. This is one approach, the other approach that we can take is initialization of some services manually, and then register those as part of the service collection initialization.

### DIFF
--- a/scripts/vscode.patch
+++ b/scripts/vscode.patch
@@ -319,10 +319,19 @@ index 23d547570e9..31dfb4fd8d4 100644
  
  export * from 'vs/editor/editor.api';
 diff --git a/src/vs/editor/standalone/browser/standaloneServices.ts b/src/vs/editor/standalone/browser/standaloneServices.ts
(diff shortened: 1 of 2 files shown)


CGNonofr:
Good job!

As an alternative, to be even more close to vscode, you can even remove the editor feature instantiations from the service initialization, and import vs/workbench/contrib/codeEditor/browser/editorFeatures.ts from monaco.ts

CompuIves:
Good one! Updated the PR.

CGNonofr (approved):

PR #236, opened 2023-11-09
Improve vite optimizeDeps config

CGNonofr:
Prevent vite from reloading the demo 2 times in dev mode after the vite cache is cleaned


kaisalmen (approved):
It works as intended!

PR #238, opened 2023-11-10
feat: add extension gallery

CompuIves:
I think this is pretty ready for an early review now. This PR enables the extensions gallery service, and all the services it requires to function. It allows you to install extensions from the extension marketplace (https://open-vsx.org/) either in web only, or on the remote if a remote is connected.

To enable this, I made some changes:

- Downgrade `semver` to v5.5.0, which is used within VSCode. This way we can stub their module.
- Allow for passing a custom product configuration to the client & the server, this allows us to define a different extension marketplace (that is not from Microsoft)
- Imported all extra required services, including their stubs


rollup/rollup.config.ts
```diff
@@ -550,6 +552,26 @@ export default (args: Record<string, string>): rollup.RollupOptions[] => {
...
+          return ''
+        }
+
+        let relativePath = path.relative(path.dirname(chunk.facadeModuleId), ASSETS_PATH)
+        if (!relativePath.startsWith('.')) {
+          relativePath = './' + relativePath
+        }
+
+        return `import "${relativePath}";\n`
+      },
```
CGNonofr, line 574: should be fixed by https://github.com/CodinGame/monaco-vscode-api/pull/241

CompuIves: Thank you!! I was fighting for a while with this.

src/service-override/extensionGallery.ts
```diff
@@ -0,0 +1,72 @@
...
+    [IExtensionManagementServerService.toString()]: new SyncDescriptor(ExtensionManagementServerService, [], true),
+    [IExtensionRecommendationsService.toString()]: new SyncDescriptor(ExtensionRecommendationsService, [], true),
+    [IExtensionRecommendationNotificationService.toString()]: new SyncDescriptor(ExtensionRecommendationNotificationService, [], true),
+    [IWebExtensionsScannerService.toString()]: new SyncDescriptor(WebExtensionsScannerService, [], true),
+    [IExtensionIgnoredRecommendationsService.toString()]: new SyncDescriptor(ExtensionIgnoredRecommendationsService, [], true),
+    [IIgnoredExtensionsManagementService.toString()]: new SyncDescriptor(IgnoredExtensionsManagementService, [], true),
+    [IExtensionManifestPropertiesService.toString()]: new SyncDescriptor(ExtensionManifestPropertiesService, [], true),
+    [IExtensionManagementService.toString()]: options.webOnly
+      ? new SyncDescriptor(WebExtensionManagementService, [], true)
+      : new SyncDescriptor(ExtensionManagementService, [], true),
```
CGNonofr, line 64: I'm a little lost here, it seems in vscode, `WebExtensionManagementService` seems to only be used inside `ExtensionManagementServerService`

`IExtensionManagementService` (via `IWorkbenchExtensionManagementService`) seems to only be bound to `ExtensionManagementService` in vscode web

is it a `hack` to disable running extension on the server when the server connection is available?

CGNonofr (changes requested):
Impressive work, well done!

Is there a way for extensions to be kept after a reload?

There are networks errors, and not working images:
image
image
do you know why?

When trying to install a remote extension in the server, I get a `FileOperationError: Unable to resolve nonexistent file '<project>/dist/extensions'`

Also, could you please simplify the git history by removing commits that are reverted later?

CompuIves:
> do you know why?

Yes, removing `res.setHeader('Cross-Origin-Resource-Policy', 'cross-origin')` makes it work, because currently we don't allow loading resources (like images) from other origins. I'm not sure if we want to remove it, though.

> When trying to install a remote extension in the server, I get a FileOperationError: Unable to resolve nonexistent file '<project>/dist/extensions'

Argh, I thought I had fixed it with [this line](https://github.com/CodinGame/monaco-vscode-api/pull/238/files#diff-185b37c3ae3995028fef9d6feb98cc0b8e3766b9710bd24fd97c8f1442179d88R22), but maybe I got the directory wrong or we need more. The explanation is this: VSCode tries to look for system extensions (which are extensions preinstalled on the system, can only be disabled and not uninstalled) in a specific directory. If that directory does not exist, it will just error out and things will be broken. The directory it looks for is in the directory where the `server.js` got executed from (or so I thought). It could perhaps be that this directory resolution changed after the asset registration change.

> Is there a way for extensions to be kept after a reload?

I think our storage service work might've fixed this! I'll rebase soon.

> Also, could you please simplify the git history by removing commits that are reverted later?

Yep!

.vscode/settings.json
```diff
@@ -2,6 +2,9 @@
   "eslint.codeAction.showDocumentation": {
     "enable": true
   },
+  "[typescript]": {
+    "editor.defaultFormatter": null
```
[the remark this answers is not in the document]
CompuIves: This is a personal preference that I can remove. I have my VSCode configured to format on save, but that messes up formatting. I found another way though, so I'll remove this.

package.json
```diff
@@ -87,7 +87,7 @@
     "rollup-plugin-copy": "^3.5.0",
     "rollup-plugin-dts": "^6.1.0",
     "rollup-plugin-styles": "^4.0.0",
-    "semver": "^7.5.4",
+    "semver": "=5.5.0",
```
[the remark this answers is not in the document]
CompuIves: The minified version of VSCode is based on semver v5.5.0. I think we can use it directly but I could use some help in configuring this in Rollup.

rollup/rollup.server.config.ts
```diff
@@ -14,6 +15,12 @@ const EXTENSIONS = ['', '.ts', '.js']
 const BASE_DIR = path.resolve(__dirname, '..')
 const TSCONFIG = path.resolve(BASE_DIR, 'tsconfig.rollup.json')
 
+const dir = 'dist/server'
+
+// We need to create an empty "extensions" folder so the vscode server can use that
+// directory to search for system extensions
+await fs.mkdir(path.join(dir, 'extensions'), { recursive: true })
```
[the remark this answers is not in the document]
CompuIves: Good one! I'll have to take a look at this again anyways because I think the directory changed.

CompuIves:
> Also, could you please simplify the git history by removing commits that are reverted later?

Will do! I'll squash everything together and separate the demo changes in another commit.

package.json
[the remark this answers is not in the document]
CGNonofr: Why doesn't VScode use semver v5.5.0 as a dependency then? What's the difference?

src/service-override/extensionGallery.ts
CGNonofr: I do think it's an interesting feature, I'm just not sure about the current implementation. Is there an issue with my suggestion?

CGNonofr:
> Yes, removing res.setHeader('Cross-Origin-Resource-Policy', 'cross-origin') makes it work, because currently we don't allow loading resources (like images) from other origins. I'm not sure if we want to remove it, though.

I'm not 100% familiar with those http headers, can you elaborate on what it does exactly and why removing it resolves the error ?

CompuIves:
> I'm not 100% familiar with those http headers, can you elaborate on what it does exactly and why removing it resolves the error ?

Actually... I think I know why it works in VSCode. The `ExtensionResourceLoaderService` has this code when resolving a resource:

```ts
		const requestInit: RequestInit = {};
		if (this.isExtensionGalleryResource(uri)) {
			requestInit.headers = await this.getExtensionGalleryRequestHeaders();
			requestInit.mode = 'cors'; /* set mode to cors so that above headers are always passed */
		}

```

And our `SimpleExtensionResourceLoaderService` doesn't have the `cors` adding yet. I'm adding that now.

CGNonofr:
> > I'm not 100% familiar with those http headers, can you elaborate on what it does exactly and why removing it resolves the error ?
> 
> Actually... I think I know why it works in VSCode. The `ExtensionResourceLoaderService` has this code when resolving a resource:
> 
> ```ts
> 		const requestInit: RequestInit = {};
> 		if (this.isExtensionGalleryResource(uri)) {
> 			requestInit.headers = await this.getExtensionGalleryRequestHeaders();
> 			requestInit.mode = 'cors'; /* set mode to cors so that above headers are always passed */
> 		}
> ```
> 
> And our `SimpleExtensionResourceLoaderService` doesn't have the `cors` adding yet. I'm adding that now.

Well spotted!

I think now `ExtensionResourceLoaderService` without customization should be enough

CompuIves:
> I think now ExtensionResourceLoaderService without customization should be enough

I was initially thinking that, but the browser version is not exported. So I can do two things:

1. Copy the code over (which I've done in the last commit)
2. Add to the VSCode diff to export the service

Disadvantage of copying is that we won't keep up with VSCode changes. I can make it an export, what do you think?

src/server/server-assets.ts
```diff
@@ -1,6 +1,7 @@
 import { registerAssets } from '../assets'
 
 registerAssets({
+  '': new URL('.', import.meta.url).toString(),
```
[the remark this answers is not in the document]
CompuIves: Works without it, I removed it.

rollup/rollup.server.config.ts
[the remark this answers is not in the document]
CompuIves: After my latest changes to extension loading, it started working properly.

CGNonofr:
> > I think now ExtensionResourceLoaderService without customization should be enough
> 
> I was initially thinking that, but the browser version is not exported. So I can do two things:
> 
> 1. Copy the code over (which I've done in the last commit)
> 2. Add to the VSCode diff to export the service
> 
> Disadvantage of copying is that we won't keep up with VSCode changes. I can make it an export, what do you think?

That's what the patch is for, we already did it plenty of times already

CompuIves:
Okay! [This](https://github.com/CodinGame/monaco-vscode-api/pull/238/commits/ff792636e8995a15f541c35fda9c7cc034a35e52 is a pretty important commit, I could use your review on it.

I noticed that extensions would start disabled, when loading VSCode. I took a look at the `extensionService`, and they have some enable logic when the extension service boots. I copied that code, instead of using the service directly, because I believe our service has some custom logic for inline extensions. I tested this code both with the extension gallery override and without, and it still works!

src/service-override/extensionGallery.ts
CompuIves: Added this!

CompuIves:
Okay, applied that change

package.json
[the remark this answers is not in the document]
CompuIves: Looks like they decided to use this UMD bundle to reduce bundle size. We can use the same if we want to!

CompuIves:
Cleaned up the commits now as well

package.json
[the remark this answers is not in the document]
CompuIves: We could try to use the same UMD bundle, but I'm not sure how we can load such a module with our rollup configuration.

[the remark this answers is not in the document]
CGNonofr: Hum not nice!

I see 2 solutions:
- trying to load the file used by vscode by transforming it via a similar approach to rollup commonjs plugin
- Install semver as an alias to not conflict with semver used in the release script and redirect to it

src/service-override/extensionGallery.ts
```diff
@@ -0,0 +1,92 @@
...
+import { IRemoteUserDataProfilesService, RemoteUserDataProfilesService } from 'vs/workbench/services/userDataProfile/common/remoteUserDataProfiles'
+import { ExtensionEnablementService } from 'vs/workbench/services/extensionManagement/browser/extensionEnablementService'
+import 'vs/workbench/services/extensionManagement/browser/extensionBisect'
+import { changeUrlDomain } from './tools/url'
+import { registerAssets } from '../assets'
+import { IInstantiationService, ILabelService, IRemoteAgentService } from '../services'
+
+registerAssets({
+  'vs/workbench/services/extensions/worker/webWorkerExtensionHostIframe.html': () => changeUrlDomain(new URL('../assets/webWorkerExtensionHostIframe.html', import.meta.url).href, undefined)
+})
```
CGNonofr, line 39: I don't think you need it here, do you?

CompuIves: You're completely right! I removed it.

package.json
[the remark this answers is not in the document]
CompuIves: I went with the second option now, that one seems less complicated to me, in terms of managing build configuration.

src/server/server.ts
```diff
@@ -1,127 +1,19 @@
...
-function parseRange (strRange: string) {
-  const match = strRange.match(/^(\d+)-(\d+)$/)
-  if (match != null) {
-    const start = parseInt(match[1]!, 10); const end = parseInt(match[2]!, 10)
-    if (start > 0 && start <= end && end <= 65535) {
-      return { start, end }
-    }
+// Initialize the product information for the server, including the extension gallery URL.
+// @ts-ignore
+globalThis._VSCODE_PRODUCT_JSON = {
```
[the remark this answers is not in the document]
CGNonofr: (Do you see that comment?)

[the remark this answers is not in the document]
CompuIves: Yep! I will apply this now.

demo/src/setup.ts
```diff
@@ -111,7 +113,20 @@ await initializeMonacoService({
...
+    productConfiguration: {
+      extensionsGallery: {
+        serviceUrl: 'https://open-vsx.org/vscode/gallery',
+        itemUrl: 'https://open-vsx.org/vscode/item',
+        resourceUrlTemplate: 'https://open-vsx.org/vscode/unpkg/{publisher}/{name}/{version}/{path}',
+        controlUrl: '',
+        nlsBaseUrl: '',
+        publisherUrl: ''
+      }
+    }
```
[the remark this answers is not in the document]
CompuIves: Waiting for https://github.com/CodinGame/monaco-vscode-api/pull/253 to apply it!

src/server/server.ts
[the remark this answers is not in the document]
CompuIves: Done!

src/service-override/environment.ts
```diff
@@ -21,7 +21,21 @@ function getServiceOverride (): IEditorOverrideServices
...
+      nameShort: 'Code - OSS Dev',
+      nameLong: 'Code - OSS Dev',
+      applicationName: 'code-oss',
+      dataFolderName: '.vscode-oss',
+      urlProtocol: 'code-oss',
+      reportIssueUrl: 'https://github.com/microsoft/vscode/issues/new',
+      licenseName: 'MIT',
+      licenseUrl: 'https://github.com/microsoft/vscode/blob/main/LICENSE.txt',
+      serverApplicationName: 'code-server-oss',
+      ...(options.productConfiguration ?? {})
```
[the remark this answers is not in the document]
CompuIves: Updated!

src/server/server.ts
```diff
@@ -1,127 +1,24 @@
-import './server-assets'
-import type { IServerAPI } from 'vs/server/node/remoteExtensionHostAgentServer'
-import { createServer } from 'vs/server/node/server.main.js'
-import { buildHelpMessage, buildVersionMessage, parseArgs } from 'vs/platform/environment/node/argv'
-import { ServerParsedArgs, serverOptions } from 'vs/server/node/serverEnvironmentService'
-import product from 'vs/platform/product/common/product'
-import { AddressInfo, ListenOptions, Socket } from 'net'
-import http from 'http'
+import { IProductService } from '../services'
```
CGNonofr, line 3: please import from the module defining it directly

services.ts should not be included by the server

src/server/server.ts
```diff
@@ -1,127 +1,24 @@
...
-    const start = parseInt(match[1]!, 10); const end = parseInt(match[2]!, 10)
-    if (start > 0 && start <= end && end <= 65535) {
-      return { start, end }
-    }
+// Initialize the product information for the server, including the extension gallery URL.
+thisWithVSCodeParams._VSCODE_PRODUCT_JSON = {
+  extensionsGallery: {
+    serviceUrl: 'https://open-vsx.org/vscode/gallery',
+    itemUrl: 'https://open-vsx.org/vscode/item',
+    resourceUrlTemplate: 'https://open-vsx.org/vscode/unpkg/{publisher}/{name}/{version}/{path}',
```
CGNonofr, line 15: So we can't override it on the server?

src/server/server.ts
```diff
@@ -1,127 +1,24 @@
...
-      process.exit(1)
-    } else {
-      console.warn(`--port "${strPort}" is not a valid number or range. Ranges must be in the form 'from-to' with 'from' an integer larger than 0 and not larger than 'end'.`)
-      process.exit(1)
-    }
-  }
-  return 8000
+// @ts-ignore
+thisWithVSCodeParams._VSCODE_PACKAGE_JSON = {
+  version: '1.83.0'
```
CGNonofr, line 23: What about injecting it? like in the ProductService

src/override/vs/base/common/semver/semver.ts
```diff
@@ -1,5 +1,2 @@
-export const gte = (): boolean => false
-export const gt = (): boolean => false
-export const lt = (): boolean => false
-export const lte = (): boolean => false
-export const valid = (): boolean => false
+// @ts-ignore
```
CGNonofr, line 1: you can add `"@types/vscode-semver": "npm:@types/semver@=5.5.0",` in devDependencies instead

src/server/server.ts
```diff
@@ -1,127 +1,24 @@
...
-      // Remote-SSH extension relies on this exact port error message, treat as an API
-      console.warn(`--port: Could not find free port in range: ${range.start} - ${range.end} (inclusive).`)
-      process.exit(1)
-    } else {
-      console.warn(`--port "${strPort}" is not a valid number or range. Ranges must be in the form 'from-to' with 'from' an integer larger than 0 and not larger than 'end'.`)
-      process.exit(1)
-    }
-  }
-  return 8000
+// @ts-ignore
```
CGNonofr, line 21: It doesn't seem to be required

src/service-override/extensionGallery.ts
```diff
@@ -0,0 +1,55 @@
...
+import 'vs/workbench/contrib/extensions/browser/extensions.contribution'
+import 'vs/workbench/contrib/extensions/browser/extensions.web.contribution'
+import 'vs/workbench/contrib/logs/common/logs.contribution'
+import { IRemoteUserDataProfilesService, RemoteUserDataProfilesService } from 'vs/workbench/services/userDataProfile/common/remoteUserDataProfiles'
+import { ExtensionEnablementService } from 'vs/workbench/services/extensionManagement/browser/extensionEnablementService'
+import 'vs/workbench/services/extensionManagement/browser/extensionBisect'
+
+export default function getServiceOverride (): IEditorOverrideServices {
+  return {
+    [ILocaleService.toString()]: new SyncDescriptor(WebLocaleService, [], false),
```
CGNonofr, line 67: Do we need it? it seems to only set and clear localStorage values we don't use

src/service-override/extensionGallery.ts
```diff
@@ -0,0 +1,55 @@
...
+    [IWebExtensionsScannerService.toString()]: new SyncDescriptor(WebExtensionsScannerService, [], true),
+    [IExtensionIgnoredRecommendationsService.toString()]: new SyncDescriptor(ExtensionIgnoredRecommendationsService, [], true),
+    [IIgnoredExtensionsManagementService.toString()]: new SyncDescriptor(IgnoredExtensionsManagementService, [], true),
+    [IExtensionManifestPropertiesService.toString()]: new SyncDescriptor(ExtensionManifestPropertiesService, [], true),
+    [IExtensionManagementService.toString()]: new SyncDescriptor(ExtensionManagementService, [], true),
+    [IBuiltinExtensionsScannerService.toString()]: new SyncDescriptor(BuiltinExtensionsScannerService, [], true),
+    [IWorkspaceExtensionsConfigService.toString()]: new SyncDescriptor(WorkspaceExtensionsConfigService, [], true),
+    [IRemoteExtensionsScannerService.toString()]: new SyncDescriptor(RemoteExtensionsScannerService, [], true),
+    [IExtensionTipsService.toString()]: new SyncDescriptor(ExtensionTipsService, [], true),
+    [IRemoteUserDataProfilesService.toString()]: new SyncDescriptor(RemoteUserDataProfilesService, [], true),
```
CGNonofr, line 83: is it related to extensions?

src/service-override/extensions.ts
```diff
@@ -445,8 +455,75 @@ export class SimpleExtensionService extends AbstractExtensionService implements
...
+    if (connection) {
+      connection.onDidStateChange(async (e) => {
+        if (e.type === PersistentConnectionEventType.ConnectionLost) {
+          this._remoteAuthorityResolverService._clearResolvedAuthority(remoteAuthority)
+        }
+      })
+      connection.onReconnecting(() => this._resolveAuthorityAgain())
+    }
+
+    return this._resolveExtensionsDefault()
```
CGNonofr, line 526: I have the feeling the whole VSCode implementation was copy/pasted now!

Our only need is to be able to provide a custom hostFactory and HostKindPicker. Maybe we can use the VSCode patch to split the SimpleExtensionService in 2 to make it take them as parameters of the constructor. That would allow to better follow VSCode changes

demo/vite.config.ts
```diff
@@ -18,7 +18,7 @@ export default defineConfig({
       apply: 'serve',
       configureServer: server => {
         server.middlewares.use((_req, res, next) => {
-          res.setHeader('Cross-Origin-Embedder-Policy', 'require-corp')
+          res.setHeader('Cross-Origin-Embedder-Policy', 'credentialless')
```
CGNonofr, line 21: Hum not sure what are the implications of that, help doesn't help but maybe I'm just tired ^^

src/server/server.ts
```diff
@@ -21,4 +23,10 @@ thisWithVSCodeParams._VSCODE_PACKAGE_JSON = {
   version: '1.83.0'
 }
 
+// Create a directory for system extensions to be installed in, VSCode
+// will try to read this folder to find system extensions, and will
+// error if it does not exist.
+const currentDirUrl = new URL('.', import.meta.url)
```
CGNonofr, line 29: nitpicking: you can probably use 
```typecript
import url from 'url'
url.fileURLToPath(import.meta.url)
```

src/service-override/extensionGallery.ts
```diff
@@ -30,14 +30,45 @@ import 'vs/workbench/contrib/logs/common/logs.contribution'
...
+
+export interface ExtensionGalleryOptions {
+  /**
+   * Whether we should only allow for web extensions to be installed, this is generally
+   * true if there is no server part.
+   */
+  webOnly: boolean
+}
+
+export default function getServiceOverride (options: ExtensionGalleryOptions = { webOnly: false }): IEditorOverrideServices {
```
CGNonofr, line 65: Can we make the parameter optional with a default value compliant to the VSCode behavior?

src/service-override/files.ts
```diff
@@ -469,7 +469,7 @@ const userDataFileSystemProvider = new InMemoryFileSystemProvider()
 void userDataFileSystemProvider.mkdir(URI.from({ scheme: Schemas.vscodeUserData, path: '/User/' }))
 
 const providers: Record<string, IFileSystemProvider> = {
-  extension: extensionFileSystemProvider,
+  'extension-fs': extensionFileSystemProvider,
```
CGNonofr, line 472: it probably needs to be changed in extensions.ts in `registerExtension` as well

CGNonofr (commented):
It seems you have eslint warning as well :)

src/service-override/files.ts
CompuIves: I checked that, but I _think_ that's not needed as it uses the filesystem directly; https://github.com/CompuIves/monaco-vscode-api/blob/25fbe41459c2fc3cdb4a1594060b6cecb5cce51e/src/service-override/files.ts#L513-L515

src/service-override/extensionGallery.ts
CompuIves: It's optional as it has an initializer:

image

src/service-override/extensionGallery.ts
CompuIves: Yes, it's required to know which extension to load from the server.

src/service-override/extensionGallery.ts
CompuIves: We don't need it!

src/service-override/extensionGallery.ts
CGNonofr: Ok I need to sleep more, my bad

src/service-override/files.ts
CGNonofr: I mean here: https://github.com/CompuIves/monaco-vscode-api/blob/25fbe41459c2fc3cdb4a1594060b6cecb5cce51e/src/extensions.ts#L100

src/server/server.ts
CompuIves: Oh interesting... I thought I had done that but maybe it got reverted or something.

src/server/server.ts
CompuIves: No, it's loaded as a side effect so this code has to load first.

CompuIves: In VSCode it's injected during build.

CGNonofr: Can't we allow to override it using a cli parameter or something?

src/service-override/extensions.ts
CompuIves: > I have the feeling the whole VSCode implementation was copy/pasted now!

Yes indeed, it's practically copy pasted 😄 . It was the best way to align the logic with VSCode when it comes to loading & enabling extensions + handling those edge cases.

So we wouldn't use the `SimpleExtensionService` anymore, but we'd pass in a parameter to the constructor that does the custom `hostFactory` and `HostKindPicker` logic?

src/service-override/files.ts
CompuIves: Oh good shout! Fixed it.

src/service-override/extensions.ts
CompuIves: I'm not very familiar with how the custom extension service works, so I'm wondering if we should do that refactor in a separate PR or this one.

CompuIves:
Fixed the lints!

src/override/vs/base/common/semver/semver.ts
CompuIves: Smart!

src/server/server.ts
CGNonofr: What about an env variable which contains an optional product.json file to load then?

src/service-override/files.ts
CGNonofr: probably nitpicking, but maybe `extension-fs` is not a so good descriptive name, since `extension` scheme will also be a fs to store extensions.

Also, I'm not sure to understand how it's supposed to work, what will use the `extension` fs provider? I don't see any registration of it, neither in that library or in VSCode itself

src/server/server.ts
CompuIves: Oh huh, interesting... I thought I had built that. I think I deleted a commit in the history because that's what I exactly did! I will update the PR.

CompuIves: Updated the PR with custom env var for giving product.json

src/service-override/extensions.ts
CGNonofr: I've added 2 commits to your PR, feel free to rewrite it as you wish

src/server/server.ts
CGNonofr: You need to add the `@rollup/plugin-replace` plugin in the rollup config for the server then

CGNonofr: or maybe instead of just setting the version, you can import the VSCode whole package.json file

src/server/server.ts
CGNonofr: just test it, and it crashed, you probably need to apply `path.dirname(` on it

src/service-override/files.ts
CGNonofr: I've just tried keeping `extension` scheme and everything seems to work fine, what is the issue?

src/server/server.ts
```diff
@@ -1,127 +1,38 @@
...
+thisWithVSCodeParams._VSCODE_PRODUCT_JSON = {
+  extensionsGallery: {
+    serviceUrl: 'https://open-vsx.org/vscode/gallery',
+    itemUrl: 'https://open-vsx.org/vscode/item',
+    resourceUrlTemplate: 'https://open-vsx.org/vscode/unpkg/{publisher}/{name}/{version}/{path}',
+    controlUrl: '',
+    nlsBaseUrl: '',
+    publisherUrl: ''
   }
-  return undefined
```
CGNonofr, line 65: for the client to be able to load server extensions files, some properties of that object are mandatory:
- `quality` (with the same value as the client - `oss`)
- `commit` should be set to `VSCODE_REF`
- `webEndpointUrlTemplate` which is used to check if the client is a valid URL to add the `Access-Control-Allow-Origin` header, it need to have the same origin as the client, so it'll probably need to be a parameter :shrug:

CGNonofr:
@CompuIves I've addressed myself all my concern, I'd like to know what you think about the last few commits

CompuIves:
Wow, thanks a lot! I'll do a review now.

src/service-override/files.ts
```diff
@@ -517,7 +517,7 @@ const userDataFileSystemProvider = new InMemoryFileSystemProvider()
 void userDataFileSystemProvider.mkdir(URI.from({ scheme: Schemas.vscodeUserData, path: '/User/' }))
 
 const providers: Record<string, IFileSystemProvider> = {
-  'extension-fs': extensionFileSystemProvider,
+  extension: extensionFileSystemProvider,
```
CompuIves: Should we go back to `extension` here?

src/service-override/extensionGallery.ts
```diff
@@ -22,13 +22,11 @@ import { RemoteExtensionsScannerService } from 'vs/workbench/services/remote/com
 import { ExtensionRecommendationNotificationService } from 'vs/workbench/contrib/extensions/browser/extensionRecommendationNotificationService'
 import { ExtensionTipsService } from 'vs/platform/extensionManagement/common/extensionTipsService'
 import { ExtensionManagementService } from 'vs/workbench/services/extensionManagement/common/extensionManagementService'
-import 'vs/workbench/contrib/extensions/browser/extensions.contribution'
-import 'vs/workbench/contrib/extensions/browser/extensions.web.contribution'
-import 'vs/workbench/contrib/logs/common/logs.contribution'
 import { IRemoteUserDataProfilesService, RemoteUserDataProfilesService } from 'vs/workbench/services/userDataProfile/common/remoteUserDataProfiles'
 import { ExtensionEnablementService } from 'vs/workbench/services/extensionManagement/browser/extensionEnablementService'
-import 'vs/workbench/services/extensionManagement/browser/extensionBisect'
```
CompuIves: Last time I noticed we need the bisect, but I assume that it's now contributed in another way?

demo/src/setup.ts
```diff
@@ -85,7 +85,12 @@ await Promise.all([
     folders: [{
       path: '/tmp'
     }]
-  }))
+  })),
+  initFile(monaco.Uri.file('/tmp/.vscode/extensions.json'), `{
```
CompuIves, line 89: ❤️

src/server/server.ts
```diff
@@ -32,7 +32,7 @@ if (PRODUCT_JSON_PATH != null) {
 // Create a directory for system extensions to be installed in, VSCode
 // will try to read this folder to find system extensions, and will
 // error if it does not exist.
-const currentDirPath = fileURLToPath(import.meta.url)
+const currentDirPath = path.dirname(fileURLToPath(import.meta.url))
```
CompuIves, line 42: Ah, apologies 🤦

CompuIves (commented):
Some small questions/notes. For the rest it looks good!

src/service-override/files.ts
CompuIves: Or rather, why are we going back to `extension`?

CGNonofr: Why did we change in the first place?

src/service-override/extensionGallery.ts
CGNonofr: Oh it seems it'seven imported by missing-services.ts, it may be an issue

CGNonofr: Ok the risk seems to only be to include some actions and make the bundle bigger for nothing if the extension gallery service override is not used

src/service-override/files.ts
CompuIves: If we use `extension`, we're conflicting with the input of the gallery, and because of this we'll see breadcrumbs for extension views:

Screenshot 2023-11-21 at 23 59 23

CGNonofr: well spotted, replaced by 3debbf8e0a97908c509eb6aaf64a23544455984f then

CGNonofr:
wrongly rebased, no conflict anymore!

CGNonofr (approved):

CompuIves:
Awesome to see this!!!  Congrats on the release! 🎉

CGNonofr:
> Awesome to see this!!! Congrats on the release! 🎉

There are still a few issues, and the demo failed to deploy, fix incoming!

CGNonofr:
Can you check https://github.com/CodinGame/monaco-vscode-api/pull/256 ? :)

PR #240, opened 2023-11-14
feat: expose activity services

CompuIves:
This PR exposes some internal interfaces for activity services. I'm currently working on using the sidebar of VSCode, while still rendering our own activitybar. To know which view is currently opened, and which views are available, I _believe_ I will need these interfaces.

### DIFF
--- a/src/service-override/views.ts
+++ b/src/service-override/views.ts
@@ -739,5 +739,9 @@ export {
   SidebarPart,
   ActivitybarPart,
   PanelPart,
-  Parts
+  Parts,
+
+  IActivityService,
+  IViewDescriptorService,
+  IPaneCompositePartService
 }


src/service-override/views.ts
```diff
@@ -739,5 +739,9 @@ export {
   SidebarPart,
   ActivitybarPart,
   PanelPart,
-  Parts
+  Parts,
+
+  IActivityService,
+  IViewDescriptorService,
+  IPaneCompositePartService
```
CGNonofr, line 746: We prefer exporting service identifiers from `services.ts` to be able to use a custom implementation without pulling vscode implementations (and not use it)

CompuIves: Sounds good, I'll put it there!

CompuIves: Noticed just now that they are in `services` 🤦 . Looks like my editor couldn't find them. I'll close this PR.

PR #241, opened 2023-11-14
Improve asset management

CGNonofr:
if FileAccess is used before `assets.ts` is imported, it crashes before monacoRequire is not defined

so move the code to the vscode patch instead


kaisalmen (commented):
Are you sure changing the assets handling doesn't have side effects?

kaisalmen (approved):
Nevermind. Got it. Overlook something in the patch,

CGNonofr:
> Nevermind. Got it. Overlook something in the patch,

Yep the patch diff is bigger than you can expect because to be able to use `import.meta.url` I need to change monaco build to accept es2020 and disable its amd build

PR #242, opened 2023-11-14
feat: expose `WorkspaceService`

CompuIves:
We sometimes want to reinitialize the workspace (e.g. when the user decides to change their folders). For this, we need to run:

```ts
const workspaceService = accessor.get(IWorkspaceContextService) as WorkspaceService
workspaceService.initialize(...)
```

So we can call `initialize` on it. To be able to do this, we need to be able to access the `WorkspaceService` type.

In this case I decided to expose it from `configuration.ts`, since it's only used when this service overridden. But let me know if you'd prefer to expose this from services!

### DIFF
--- a/src/service-override/configuration.ts
+++ b/src/service-override/configuration.ts
@@ -124,6 +124,11 @@ registerServiceInitializePreParticipant(async (accessor) => {
 
 const MemoizedInjectedConfigurationService = memoizedConstructor(InjectedConfigurationService)
 
+export async function reinitializeWorkspace (workspace: IAnyWorkspaceIdentifier): Promise<void> {
+  const workspaceService = StandaloneServices.get(IWorkspaceContextService) as WorkspaceService
+  await workspaceService.initialize(workspace)
+}
+
 export default function getServiceOverride (defaultWorkspace: URI | IAnyWorkspaceIdentifier): IEditorOverrideServices {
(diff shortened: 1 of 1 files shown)


CGNonofr:
You try to expose an impl here, so we mustn't export it from a common file

Ok I didn't realize it's possible to call initialize more than once! I've tried it and it seems to work, also the code make me think it's allowed

I'm just a little confused, because the only place where it's called inside VSCode is https://github.com/microsoft/vscode/blob/54821ee1f14beca4866abd7de86175b4794b030d/src/vs/workbench/services/workspaces/browser/workspaceEditingService.ts#L51, and the window is reload just after

I'm ok to expose the class though, but maybe exposing a `reinitializeWorkspace` is more practical

CompuIves:
I still need to test if this works properly, though. Like you said, VSCode does not do it. I do see that it has the logic to remove/add folders through this flow, though. So I expect it to work.

CGNonofr:
note: in the demo, we use the `vscode.workspace.updateWorkspaceFolders` to update the workspace folder. It requires the current workspace to be a workspace file though

CompuIves:
One of the main goals with this update is to update the `id` of the workspace. I checked, and it does not seem possible with `updateWorkspaceFolders`.

CGNonofr (approved):

PR #243, opened 2023-11-14
feat: allow specifying editor restoration

CompuIves:
This allows us to configure whether the editor should restore its view state from the last load, by reading from the `StorageService`.


CGNonofr (approved):
Looks good

CompuIves:
Oh... Actually I missed something. We filter out `registerEditorSerializer` contributions, so the editors are not serialized. Let me try again!

PR #246, opened 2023-11-14
Fix editor.addAction keybindings

### LINKED ISSUE #244: Monaco actions didn't works with keybinding service override
Hi Team!

I have a lot of custom actions for monaco-editor, they all registered after initialisation process via `createConfiguredEditor` call.
And after that al actions didn't works, just ignoring by editor. What i do wrong?

CGNonofr:
ctrl+K in the demo doesn't work anymore

Not sure about what caused the regression, don't know how it was working

fix #244


kaisalmen (approved):
LGTM

PR #249, opened 2023-11-16
feat: expose additional options for registering views

CompuIves:
I'm preparing us for phase 2 of the implementation of VSCode: replacing our sidebar with the one from VSCode.

To enable this, we'll need to render some of our React components inside the sidebar, sometimes as a new view item (separate view in activity bar). With this PR, I'm extending the capabilities of `registerCustomView` to allow for:

1. More finegrained options (in terms of order, collapsed, etc...)
2. Ability to register a new item in existing views (like the explorer), instead of creating a new view

Let me know if this change makes sense to you!

### DIFF
--- a/src/service-override/views.ts
+++ b/src/service-override/views.ts
@@ -1,6 +1,6 @@
 import { IEditorOverrideServices, StandaloneServices } from 'vs/editor/standalone/browser/standaloneServices'
 import { SyncDescriptor } from 'vs/platform/instantiation/common/descriptors'
-import { IViewContainersRegistry, IViewDescriptor, IViewDescriptorService, IViewsRegistry, IViewsService, ViewContainerLocation, Extensions as ViewExtensions } from 'vs/workbench/common/views'
+import { IViewContainersRegistry, IViewDescriptor, IViewDescriptorService, IViewsRegistry, IViewsService, ViewContainer, ViewContainerLocation, Extensions as ViewExtensions } from 'vs/workbench/common/views'
 import { ViewsService } from 'vs/workbench/browser/parts/views/viewsService'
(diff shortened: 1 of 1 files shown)


src/service-override/views.ts
```diff
@@ -388,12 +388,16 @@ interface CustomViewOption {
...
+  canToggleVisibility?: boolean
+  hideByDefault?: boolean
+  collapsed?: boolean
 }
 
 function registerCustomView (options: CustomViewOption): IDisposable {
   const iconUrl = options.icon != null ? URI.parse(options.icon) : undefined
 
-  const VIEW_CONTAINER = Registry.as<IViewContainersRegistry>(ViewExtensions.ViewContainersRegistry).registerViewContainer({
+  const VIEW_CONTAINER = options.viewContainer ?? Registry.as<IViewContainersRegistry>(ViewExtensions.ViewContainersRegistry).registerViewContainer({
```
CGNonofr, line 400: How did you create the view container?

CGNonofr: btw, you may want to NOT deregistering the view container then line 456

CompuIves: Good one. I get them through `IViewDescriptorService`. I run this:

```
viewDescriptorService
            .getViewContainersByLocation(0)
```

Where 0 corresponds to the sidebar.

CompuIves:
Updated the disposal!

src/service-override/views.ts
CGNonofr: Ok I have no idea why the variable is in upper case, do you mind changing it?

CompuIves: Sounds good, it's update!

CGNonofr (approved):

PR #250, opened 2023-11-16
fix(vsix-plugin): set mime type in dev mode

CompuIves:
In dev mode the mime type of the base64 string is hardcoded to `text/javascript`. This works for most files, but it won't work for e.g. file icons, which are svg. In that case, it will not render anything:
image

With this change, we use the inferred mime type for the base64. Now the main question is... Will this work for all file types?

### DIFF
--- a/src/rollup-vsix-plugin.ts
+++ b/src/rollup-vsix-plugin.ts
@@ -101,7 +101,8 @@ export default function plugin ({
         const assetPath = getVsixPath(resource.realPath ?? resource.path)
         let url: string
         if (process.env.NODE_ENV === 'development') {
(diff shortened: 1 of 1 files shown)


CGNonofr:
Oh well spotted!

What is the issue though? all that mimeType stuff was added because it was important for `vite` to set the `accept` header, otherwise it was trying to bundle/compile asset files. But for base64 urls, I'm not sure what can be the issue

CompuIves:
The main issue is that VSCode injects these base64 URLs as inline images in CSS. If the mimetype is not right, the browser refuses to render the icon.

CGNonofr (approved):

CompuIves:
Would it be possible to cut a new release with these changes? That would enable me to do some additional work on the repo where we use the library 😄

PR #251, opened 2023-11-17
Log service override

CGNonofr:
Extract it from OutputService

also a few related fixes/changes


CompuIves:
Nice PR! Does this also allow us to use more exports from VSCode directly?

src/service-override/environment.ts
```diff
@@ -20,8 +20,7 @@ function getServiceOverride (): IEditorOverrideServices
 
 function getServiceOverride (options?: IWorkbenchConstructionOptions): IEditorOverrideServices {
   return {
-    [IEnvironmentService.toString()]: new SyncDescriptor(InjectedBrowserWorkbenchEnvironmentService, [], true),
-    [IBrowserWorkbenchEnvironmentService.toString()]: new SyncDescriptor(InjectedBrowserWorkbenchEnvironmentService, [undefined, options], true)
```
CompuIves, line 24: Is this not needed anymore? I believe I use this to inject product configuration in the extensions gallery PR.

src/service-override/extensions.ts
```diff
@@ -50,11 +50,12 @@ import { IStorageService } from 'vs/platform/storage/common/storage'
 import { ILabelService } from 'vs/platform/label/common/label'
 import { ExtensionKind } from 'vs/platform/environment/common/environment'
 import { ExtensionDescriptionRegistrySnapshot } from 'vs/workbench/services/extensions/common/extensionDescriptionRegistry'
-import getOutputServiceOverride from './output'
 import { changeUrlDomain } from './tools/url'
 import { registerAssets } from '../assets'
 import { unsupported } from '../tools'
 import 'vs/workbench/api/browser/extensionHost.contribution'
+// We need to import missing service here because extHost.common.services registers ILoggerService and we want the implementation from missing-service to win
```
CompuIves, line 57: Is there still a way in this case to use the one from `extHost`? I can imagine that's used to log logs from extensions.

CompuIves (commented):
Couple of minor questions!

src/service-override/environment.ts
CGNonofr: IBrowserWorkbenchEnvironmentService is just a refined version of the other one, the toString method return the same

src/service-override/extensions.ts
CGNonofr: This file is not supposed to be used in the app but in the extHost worker instead. We include it by hands for the local extHost but we don't want it to override the services. The extHost logger service rôle is to send the message through the rpcProtocol to the app logger service

CGNonofr:
> Nice PR! Does this also allow us to use more exports from VSCode directly?

Missed that comment!

What do you mean?

CompuIves (approved):

CompuIves:
Not entirely certain, I thought something had changed in Rollup exports, but I think I was wrong. Approved it now as it looks good!

CGNonofr:
The difference in the rollup build is that previously, service-overrides that were imported by `services.ts` as default were included in monaco-vscode-api directly. Now their code is in their dedicated package and monaco-vscode-api has them as dependencies

CGNonofr:
What I was trying to fix is that, as soon as a file is used at 2 different places, it was included in monaco-vscode-api to not be duplicated in the 2 sub-libraries including it.

The problem with that is if a service-override A was including another one B, the A service-override was considered used at 2 different places and included in monaco-vscode-api. 

Now, B has A as dependency and monaco-vscode-api should be lighter

PR #253, opened 2023-11-20
fix: apply custom product configuration

CompuIves:
Currently the custom `productConfiguration` is not applied. VSCode applies it manually [here](https://github.com/CompuIves/vscode/blob/aaa0f32ee50f6836d9931d12c8cb0d6f28248bb6/src/vs/workbench/browser/web.main.ts#L253). We should do the same, but in our way :).

### DIFF
--- a/src/services.ts
+++ b/src/services.ts
@@ -5,8 +5,10 @@ import { ITextModelContentProvider } from 'vs/editor/common/services/resolverSer
 import { IColorTheme } from 'vs/platform/theme/common/themeService'
 import { StorageScope, StorageTarget } from 'vscode/src/vs/platform/storage/common/storage'
 import { IEditorOverrideServices, StandaloneServices } from 'vs/editor/standalone/browser/standaloneServices'
+import { mixin } from 'vs/base/common/objects'
 import { GetLeadingNonServiceArgs, IInstantiationService, ServiceIdentifier, ServicesAccessor } from 'vs/platform/instantiation/common/instantiation'
 import { IAction } from 'vs/base/common/actions'
+import { IProductService } from 'vs/platform/product/common/productService'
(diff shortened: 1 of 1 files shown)


src/services.ts
```diff
@@ -31,6 +32,19 @@ export async function initialize (overrides: IEditorOverrideServices, container:
   initializeWorkbench(container, configuration)
 
   const instantiationService = StandaloneServices.initialize({
+    [IProductService.toString()]: Object.assign({
```
CGNonofr, line 35: what is the difference between `Object.assign` and `mixin`?

CompuIves: No difference if there's no additional param `overwrite=false` given:

https://github.com/CompuIves/vscode/blob/master/src/vs/base/common/objects.ts#L85-L110

CompuIves: Oh, and it does in recursively...

CompuIves: Changed it to `mixin` to err on the safe side!

src/services.ts
```diff
@@ -16,6 +17,7 @@ import getFileServiceOverride from './service-override/files'
 import getQuickAccessOverride from './service-override/quickaccess'
 import { serviceInitializedBarrier, serviceInitializedEmitter, startup } from './lifecycle'
 import { initialize as initializeWorkbench } from './workbench'
+import { IProductService } from './services'
```
CGNonofr, line 20: You import it from the file itself? :)

CompuIves: 🤯

CompuIves: Fixed

CGNonofr (approved):

CGNonofr:
you have conflicts :-\

CompuIves:
Rebased!

CGNonofr (approved):

PR #257, opened 2023-11-23
feat: enable progress service

CompuIves:
This enables the progress service as part of the view API. Currently, views like search are not showing a progress bar in the UI when doing the search, because we're using the standalone service for it. With this change, it will show a progress bar for actions inside the different views.

### DIFF
--- a/src/service-override/views.ts
+++ b/src/service-override/views.ts
@@ -87,6 +87,8 @@ import { ConfirmResult } from 'vs/platform/dialogs/common/dialogs'
 import { ILayoutService } from 'vs/platform/layout/browser/layoutService'
 import { IBannerService } from 'vs/workbench/services/banner/browser/bannerService'
 import { ITitleService } from 'vs/workbench/services/title/common/titleService'
+import { IProgressService } from 'vs/platform/progress/common/progress'
+import { ProgressService } from 'vs/workbench/services/progress/browser/progressService'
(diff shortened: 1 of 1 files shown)


src/service-override/search.ts
```diff
@@ -14,6 +16,7 @@ export default function getServiceOverride (): IEditorOverrideServices {
     [ISearchService.toString()]: new SyncDescriptor(SearchService, [], true),
     [ISearchViewModelWorkbenchService.toString()]: new SyncDescriptor(SearchViewModelWorkbenchService, [], true),
     [ISearchHistoryService.toString()]: new SyncDescriptor(SearchHistoryService, [], true),
-    [IReplaceService.toString()]: new SyncDescriptor(ReplaceService, [], true)
+    [IReplaceService.toString()]: new SyncDescriptor(ReplaceService, [], true),
+    [IProgressService.toString()]: new SyncDescriptor(ProgressService, [], true)
```
CGNonofr, line 20: I'm not sure it makes sense to have it here since this service-override already required the view service override which includes it (And referencing it from both service override forces the code to be in monaco-vscode-api instead of the service override lib)

CompuIves: Ah, 🤦 . Completely true. Updating now...

CompuIves:
Updated PR!

CGNonofr (approved):

PR #258, opened 2023-11-23
fix: make sure to transform extension-file

CompuIves:
We missed this one 😄


CGNonofr (approved):

CGNonofr:
What are the visible consequences?

CompuIves:
I've no idea, actually. I initially thought it was related to something else, and noticed it while debugging.

PR #260, opened 2023-11-24
Fix dependency injection loop

### LINKED ISSUE #259: Demo: Console log suddenly full of errors
Latest main. I built it on my Ubuntu WSL as usual. And I just checked the [deployed version](https://codingame.github.io/monaco-vscode-api/) same there. Are you aware of this already?
image
image

CGNonofr:
fixes #259

VSCode PR: https://github.com/microsoft/vscode/pull/199037

### DIFF
--- a/scripts/vscode.patch
+++ b/scripts/vscode.patch
@@ -599,6 +599,22 @@ index e22cfe1b45e..b48c1babf86 100644
  export class DiskFileSystemProvider extends AbstractDiskFileSystemProvider implements
  	IFileSystemProviderWithFileReadWriteCapability,
  	IFileSystemProviderWithOpenReadWriteCloseCapability,
+diff --git a/src/vs/platform/instantiation/common/instantiationService.ts b/src/vs/platform/instantiation/common/instantiationService.ts
+index f0db479e3c1..b8d93aae3f2 100644
+--- a/src/vs/platform/instantiation/common/instantiationService.ts
++++ b/src/vs/platform/instantiation/common/instantiationService.ts
+@@ -195,8 +195,10 @@ export class InstantiationService implements IInstantiationService {
+ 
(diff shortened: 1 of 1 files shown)


scripts/vscode.patch
```diff
@@ -599,6 +599,22 @@ index e22cfe1b45e..b48c1babf86 100644
...
  	IFileSystemProviderWithOpenReadWriteCloseCapability,
+diff --git a/src/vs/platform/instantiation/common/instantiationService.ts b/src/vs/platform/instantiation/common/instantiationService.ts
+index f0db479e3c1..b8d93aae3f2 100644
+--- a/src/vs/platform/instantiation/common/instantiationService.ts
++++ b/src/vs/platform/instantiation/common/instantiationService.ts
+@@ -195,8 +195,10 @@ export class InstantiationService implements IInstantiationService {
+ 
+ 				if (instanceOrDesc instanceof SyncDescriptor) {
+ 					const d = { id: dependency.id, desc: instanceOrDesc, _trace: item._trace.branch(dependency.id, true) };
++					if (!graph.lookup(d)) {
```
CompuIves, line 610: So if the service is already in the graph, we don't add it to the stack?

CompuIves: It's interesting, because it would still "evaluate" the dependency, but not initialize it. It looks good to me.

CompuIves (approved):

scripts/vscode.patch
CGNonofr: I'm a little confused by that code, because even if I remove the adding in the stack, everything works fine :shrug:

So I tried the safest change I can imagine, we'll see the reaction of the Microsoft team on it

> So if the service is already in the graph, we don't add it to the stack?

The service being on the graph means we already compute its dependencies at that point, right?

PR #264, opened 2023-11-27
Replace map by forEach to prevent treeshaking

### LINKED ISSUE #262: EditorGroupWatermark is not shown due to tree shaker
The shortcuts that show in here:

image

Are not shown in our version of VSCode, because the code responsible for creating the shortcuts (https://github.com/microsoft/vscode/blob/1a5daa3a0231a0fbba4f14db7ec463cf99d7768e/src/vs/workbench/browser/parts/editor/editorGroupWatermark.ts#L136-L147) uses `.map()`, and that is tree shaken away in the final build (it assumes no side effects). Is there a way that we can still keep this `map()` in the final bundle?

(It's funny, because it still calculates what shortcuts should be shown. That is _not_ shaken away).

CGNonofr:
fix #262

vscode PR: https://github.com/microsoft/vscode/pull/199194

### DIFF
--- a/scripts/vscode.patch
+++ b/scripts/vscode.patch
@@ -380,46 +380,66 @@ index cf9154ce7d6..5c954f75f8e 100644
  	public static validateOptions(options: IEditorOptions): ValidatedEditorOptions {
  		const result = new ValidatedEditorOptions();
 diff --git a/src/vs/editor/contrib/find/browser/findController.ts b/src/vs/editor/contrib/find/browser/findController.ts
-index ce1cdb9084e..0836c1d3f98 100644
+index ce1cdb9084e..7083bcde458 100644
 --- a/src/vs/editor/contrib/find/browser/findController.ts
 +++ b/src/vs/editor/contrib/find/browser/findController.ts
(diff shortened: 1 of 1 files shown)


scripts/vscode.patch
```diff
@@ -379,6 +379,50 @@ index cf9154ce7d6..5c954f75f8e 100644
...
+-				selections.map(selection => {
++				selections.forEach(selection => {
+ 					if (selection.endColumn === 1 && selection.endLineNumber > selection.startLineNumber) {
+ 						selection = selection.setEndPosition(
+ 							selection.endLineNumber - 1,
+@@ -270,7 +270,7 @@ export class CommonFindController extends Disposable implements IEditorContribut
+ 						return selection;
+ 					}
+ 					return null;
+-				}).filter(element => !!element);
```
CGNonofr, line 399: I'm not sure what was attempted with that filter or if I missed something :shrug:

CGNonofr: I have the feeling they meant `selections = selections.map(selection => {`

Issue #265, opened 2023-11-28
How do I enable only problems panel tab?


levensta:
I use pure monaco-editor with its own settings and language validation. I want to add the ability to open a panel that displays a list of markers with errors, but I have not found anyone trying to solve this problem

<img width="604" alt="image" src="https://github.com/CodinGame/monaco-vscode-api/assets/64372619/acf9d01c-911e-4892-98f0-dd9a153c8304">

The `monaco-vscode-api` looks like something that might help me, but I looked at the [demo](https://github.com/CodinGame/monaco-vscode-api/tree/main/demo) in the repository and I'm afraid it's too big a package for a single feature with a panel tab.

How to properly install and configure `monaco-vscode-api` so that only the problems panel tab is used, so as not to load the bundle with extensions that won't be used?

CGNonofr:
I'm afraid that you won't be able to only display the marker panel without pulling anything else for some reasons:
- You may notice that the markers panel includes a header displayed in the tab section and the panel part, I don't know how it would work in standalone
- The panel part and services are managing the panel lifecycle (visible or not, its layout...)

What you can do is only using the `view` and `marker` service overrides and rendering only the panel part though if the bundle size is not a blocking problem

CGNonofr:
probably the view service override won't work without some other services override, like the configuration service override. It's hard to know all working combinations

PR #266, opened 2023-11-28
Improve view initialization

CGNonofr:
May be requires by #265 to be able to set the `Problems` view as default view

also includes forgotten working copy service override


kaisalmen:
@CGNonofr will test this locally today.

kaisalmen:
@CGNonofr the works as configured after I reset the user data. But that makes sense. 😊 

Where is the all the new views code coming from? You did not just write that, did you?

CGNonofr:
> Where is the all the new views code coming from? You did not just write that, did you?

Nop! it comes from the VSCode layout service but I didn't find a way to not have to copy/paste it

kaisalmen:
> Nop! it comes from the VSCode layout service but I didn't find a way to not have to copy/paste it

👍

kaisalmen (approved):
Works as expected after clearing user data.

PR #272, opened 2023-11-30
Make extension service optional

CGNonofr:
@kaisalmen it will require to add it back in monaco-languageclient

### DIFF
--- a/scripts/vscode.patch
+++ b/scripts/vscode.patch
@@ -740,18 +740,6 @@ index ccfbc3f7d9b..5fe46b464b1 100644
  	const appInsightsCore = new oneDs.AppInsightsCore();
  	const collectorChannelPlugin: PostChannel = new postPlugin.PostChannel();
  	// Configure the app insights core to send to collector++ and disable logging of debug info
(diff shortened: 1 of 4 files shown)


CGNonofr:
> @CGNonofr ok, can you tell me why you need to do this?

This library can be used for other things than using VSCode extensions, I just try to make the library as modular as possible

Do you think it's a bad idea?

CGNonofr:
> > This library can be used for other things than using VSCode extensions, I just try to make the library as modular as possible
> 
> Interesting. Do you have an example?
> 
> > Do you think it's a bad idea?
> 
> No, just want to understand the motivation 🙂

You can use the configuration/keybindings/view service override without needing anything else. It doesn't correspond to any use-case I'll have but better keep the library as modular as possible

kaisalmen:
@CGNonofr didn't you put the `fileService` there by intention in the past. Was that necessity removed after recent changes? Just want to prevent potential regressions.
Can you release a next version for testing once you are done with the changes? Thanks

CGNonofr:
> @CGNonofr didn't you put the `fileService` there by intention in the past. Was that necessity removed after recent changes? Just want to prevent potential regressions. Can you release a next version for testing once you are done with the changes? Thanks

Ok you made me change my mind :)

PR #275, opened 2023-12-01
Add missing extension points

CGNonofr:


src/missing-services.ts
```diff
@@ -1732,7 +1732,21 @@ registerSingleton(INotebookEditorModelResolverService, class NotebookEditorModel
 registerSingleton(IWorkingCopyEditorService, WorkingCopyEditorService, InstantiationType.Delayed)
 registerSingleton(IUserActivityService, UserActivityService, InstantiationType.Delayed)
 registerSingleton(ICanonicalUriService, CanonicalUriService, InstantiationType.Delayed)
-registerSingleton(IExtensionStatusBarItemService, ExtensionStatusBarItemService, InstantiationType.Delayed)
+registerSingleton(IExtensionStatusBarItemService, class ExtensionStatusBarItemService implements IExtensionStatusBarItemService {
```
CompuIves, line 1735: This is interesting, I believe we're already running extensions that contribute status bar items. Does this add something else as well?

CompuIves (approved):
Straightforward PR!

src/missing-services.ts
CGNonofr: I've just moved the implementation into the dedicated service override, no change expected here

CGNonofr: Maybe you're not talking about this specific line but the new `StatusBarItemsExtensionPoint`? then yes probably new things to expect but I don't really know why (:

PR #276, opened 2023-12-01
Split local ext host

CGNonofr:
The local ext host pulls all the code from the worker extHost into the main bundle

This PR make it optional

It's automatically registered when using `import ... from 'vscode'`

~~:warning: It relies on top-level await to not have to wait for a promise before using that api~~


kaisalmen:
@CGNonofr will look at this today

kaisalmen (approved):
Looking good and tested successfully with `monaco-languageclient`, see https://github.com/TypeFox/monaco-languageclient/pull/581

I have one question: Do you intentionally use `monaco-editor` in the demo, so you are able to test the treemending script or do you want to switch at one point to `@codingame/monaco-editor-treemended`?

CGNonofr:
> I have one question: Do you intentionally use `monaco-editor` in the demo, so you are able to test the treemending script or do you want to switch at one point to `@codingame/monaco-editor-treemended`?

It just wasn't changed when `@codingame/monaco-editor-treemended` was introduced. I'm still not sure what should be the recommended way :-\

CGNonofr:
@kaisalmen I've rebased (and added 8d78f220aabb18f5df126a4b1e6f6a540577f931) can you please approve once again?

kaisalmen:
> It just wasn't changed when @codingame/monaco-editor-treemended was introduced. I'm still not sure what should be the recommended way :-\

@CGNonofr the extra package all in all is simpler, but when upgrading from one version to the next cleaning seems to be a must (`node_modules` and eventually `package-lock.json`)

CGNonofr:
> > It just wasn't changed when @codingame/monaco-editor-treemended was introduced. I'm still not sure what should be the recommended way :-\
> 
> @CGNonofr the extra package all in all is simpler, but when upgrading from one version to the next cleaning seems to be a must (`node_modules` and eventually `package-lock.json`)

I add the feeling it was the cons of the treemending script: it's not reapplied if monaco-editor version didn't change so you need to delete the node_modules.

Why whould we need to clean node_modules and package-lock.json using monaco-editor-treemended?

kaisalmen:
> Why whould we need to clean node_modules and package-lock.json using monaco-editor-treemended?

That is some awkwardness of npm. E.g. when I upgraded from `1.83.12` to `1.83.15-next.0` this morning in `monaco-languageclient` `monaco-editor` was not consistently resolved. Only after cleaning and re-installing everything was fine again. Don't know if that is better with npm 10 / node 20

PR #278, opened 2023-12-04
feat: add extension url and authentication service

CompuIves:
We need these ones to support the "GitHub Pull Requests and Issues" extension from the marketplace (which would be insane if we could support that!)

Keep in mind that this in of itself would not be enough, the user would have to implement the GitHub authentication manually by overriding the github authentication extension manually. But that's everything needed!

### DIFF
--- a/src/service-override/extensionGallery.ts
+++ b/src/service-override/extensionGallery.ts
@@ -24,9 +24,10 @@ import { ExtensionTipsService } from 'vs/platform/extensionManagement/common/ext
 import { ExtensionManagementService } from 'vs/workbench/services/extensionManagement/common/extensionManagementService'
 import { IRemoteUserDataProfilesService, RemoteUserDataProfilesService } from 'vs/workbench/services/userDataProfile/common/remoteUserDataProfiles'
 import { ExtensionEnablementService } from 'vs/workbench/services/extensionManagement/browser/extensionEnablementService'
-import { IInstantiationService, ILabelService, IRemoteAgentService } from '../services'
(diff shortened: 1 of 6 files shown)


src/service-override/extensionGallery.ts
```diff
@@ -82,6 +85,8 @@ export default function getServiceOverride (options: ExtensionGalleryOptions = {
     [IRemoteExtensionsScannerService.toString()]: new SyncDescriptor(RemoteExtensionsScannerService, [], true),
     [IExtensionTipsService.toString()]: new SyncDescriptor(ExtensionTipsService, [], true),
     [IRemoteUserDataProfilesService.toString()]: new SyncDescriptor(RemoteUserDataProfilesService, [], true),
-    [IWorkbenchExtensionEnablementService.toString()]: new SyncDescriptor(ExtensionEnablementService, [], true)
+    [IWorkbenchExtensionEnablementService.toString()]: new SyncDescriptor(ExtensionEnablementService, [], true),
+    [IExtensionUrlHandler.toString()]: new SyncDescriptor(ExtensionUrlHandler, [], true),
+    [IAuthenticationService.toString()]: new SyncDescriptor(AuthenticationService, [], true)
```
CGNonofr, line 90: Shouldn't this one be in its own service override?

CompuIves: Yeah, I can split this one out! I was thinking it would only make sense to need this with the extension marketplace, but on second thought I'm wrong on that.

CompuIves: Moved!

PR #279, opened 2023-12-04
feat: add scm api support

CompuIves:
Tell me if I'm going too far!

This adds the SCM extension API, which allows the user to add an SCM implementation, or use the git extension directly if they have vscode-server (maybe we should add a pipeline to also compile server extensions and republish them?).

Here's the demo using a bare-bones API:

Screenshot 2023-12-04 at 22 34 53


CGNonofr:
> Tell me if I'm going too far!

It's completely fine, adding a service override is completely harmless for anyone not using it, I would be happy if this library is able to do everything VSCode does

> This adds the SCM extension API, which allows the user to add an SCM implementation, or use the git extension directly if they have vscode-server (maybe we should add a pipeline to also compile server extensions and republish them?).

I agree, but I don't know how to implement it, do you have an idea?

src/service-override/scm.ts
```diff
@@ -0,0 +1,16 @@
...
+import { IQuickDiffService } from 'vs/workbench/contrib/scm/common/quickDiff'
+import { QuickDiffService } from 'vs/workbench/contrib/scm/common/quickDiffService'
+import 'vs/workbench/contrib/scm/browser/scm.contribution'
+import { SyncDescriptor } from '../services'
+
+export default function getServiceOverride (): IEditorOverrideServices {
+  return {
+    [ISCMService.toString()]: new SyncDescriptor(SCMService, [], true),
+    [ISCMViewService.toString()]: new SyncDescriptor(SCMViewService, [], true),
+    [IQuickDiffService.toString()]: new SyncDescriptor(QuickDiffService, [], true)
```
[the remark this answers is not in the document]
CompuIves: Yes, this is registered as a singleton in `scm.contribution`. Extensions can contribute an `IQuickDiffService`.

CompuIves:
> I agree, but I don't know how to implement it, do you have an idea?

Not yet. But I can take a look at it. Will think about that separately.

src/service-override/scm.ts
```diff
@@ -0,0 +1,16 @@
+import { IEditorOverrideServices } from 'vs/editor/standalone/browser/standaloneServices'
+import { ISCMService, ISCMViewService } from 'vs/workbench/contrib/scm/common/scm'
+import { SCMService } from 'vs/workbench/contrib/scm/common/scmService'
+import { SCMViewService } from 'vs/workbench/contrib/scm/browser/scmViewService'
+import { IQuickDiffService } from 'vs/workbench/contrib/scm/common/quickDiff'
+import { QuickDiffService } from 'vs/workbench/contrib/scm/common/quickDiffService'
+import 'vs/workbench/contrib/scm/browser/scm.contribution'
+import { SyncDescriptor } from '../services'
```
CGNonofr, line 8: Can you please import it directly from the right vscode file?

CompuIves: Oh right! I keep forgetting that!

CGNonofr (approved):

PR #282, opened 2023-12-05
Testing service override

CGNonofr:
@CompuIves Do you think I'm going too far? :smile: 

Also I've just discovered https://github.com/microsoft/vscode-extension-samples/tree/main


CompuIves (commented):
Wow! This is a new API to me! Really cool, in a first glance the code makes sense, but I'd love to also run it. Most probably I can do that tomorrow morning (in about 10 hours). Will follow up!

CompuIves (approved):
Wow, this is incredible! Just used the demo and it works really well. Didn't even know this API existed!

PR #287, opened 2023-12-06
fix: enable multiple formatters and format settings

### LINKED ISSUE #286: Formatting configuration is treeshaken away
[This configuration](https://github.com/microsoft/vscode/blob/e7e037083ff4455cf320e344325dacb480062c3c/src/vs/workbench/contrib/format/browser/formatActionsMultiple.ts#L246-L258) is currently tree-shaken away, and it causes the formatting decision maker to ignore `editor.defaultFormatter`. I see that the file is imported [here](https://github.com/microsoft/vscode/blob/e7e037083ff4455cf320e344325dacb480062c3c/src/vs/workbench/contrib/format/browser/format.contribution.ts#L6), and the `format.contribution` file is imported in our contributions. So I _assume_ that it's tree-shaken away, but I don't understand why. Is it because the class is not exported?

CompuIves:
Fixes #286

### DIFF
--- a/rollup/rollup.config.ts
+++ b/rollup/rollup.config.ts
@@ -541,7 +541,8 @@ export default (args: Record<string, string>): rollup.RollupOptions[] => {
           path.includes('vs/workbench/contrib/codeEditor/browser/') ||
           path.includes('extHost.common.services') ||
           path.includes('extHost.worker.services') ||
-          path.includes('inlayHintsAccessibilty')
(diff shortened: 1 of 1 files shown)


CGNonofr (approved):

CompuIves:
Would it be possible to include this one in a release? It solves a bug that we've seen reported by some users. If it's not possible that's okay too, then I'll work on a workaround on our side!

PR #309, opened 2024-01-09
a few minor fixes

CGNonofr:
A few minor fixes

The user data sync service override is not usable for the moment at it relies on a closed-source server hosted behing `https://vscode-sync.trafficmanager.net/`. But it may be useful in the future (see https://code.visualstudio.com/docs/editor/settings-sync#_can-i-use-a-different-backend-or-service-for-settings-sync)

The welcome walkthrough relies on a `syncStatus` context key to hide or display the suggestion to sync your data, and unfortunately, without the user data sync services, the context key remains undefined, while the condition is `syncStatus != uninitialized`, so the suggestion is displayed. The only way to hide it is ironically to register the user data sync service override without configuring any server

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,10 +192,12 @@ Additionally, several packages that include the VSCode version of some services 
   - Support for the VSCode marketplace, it allows to install extensions from the marketplace.
 - **Chat**: `@codingame/monaco-vscode-chat-service-override`
   - Support for chat and inline chat features
-- **Notebook**
+- **Notebook**: `@codingame/monaco-vscode-notebook-service-override`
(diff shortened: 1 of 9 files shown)


kaisalmen (changes requested):
You forgot to update the README with info about user-data-sync-service-override 😉

CGNonofr:
> You forgot to update the README with info about user-data-sync-service-override 😉

Oops! => d12dd55fd9b525fe2616389f17fd1866332642eb

kaisalmen (approved):
Let's go!

PR #310, opened 2024-01-11
Cleanup and add missing service overrides

CGNonofr:
I don't think there is any relevant service override left to implement :)


kaisalmen:
@CGNonofr 🍾 I still need to test it

kaisalmen (approved):
Looking good! Next is update to vite 5 / rollup 4? 🤯

CGNonofr:
> Looking good! Next is update to vite 5 / rollup 4? 🤯

We are already using rollup 4, don't we?

kaisalmen:
> We are already using rollup 4, don't we?

Oh yes, nevermind! 😊 Then switching to vite 5 should be easy

CGNonofr:
> > We are already using rollup 4, don't we?
> 
> Oh yes, nevermind! 😊 Then switching to vite 5 should be easy

I've already tried, there is a weird error about experimental decorator that looks blocking but is not (but still ugly). I was hoping them to fix it but it doesn't seem there is any change since then

PR #315, opened 2024-01-13
Fixed asset paths in dev on Windows

dkattan:
In my environment import.meta.url does not maintain the relative path to the extension in node_modules:
image

So I added NPM package import-meta-resolve and used their resolve method by replacing
```js
const resolved = await import.meta.resolve!(path, url.pathToFileURL(args.path));
```
with
```js
const resolved = await resolve(path, url.pathToFileURL(args.path).toString());
```

But then I had an issue because import.meta.url starts with file://C:/ causing the pathname to be /C:/node_modules which is obviously not correct

image

This can be mitigated by replacing

```js
const pathname = new URL(req.originalUrl, import.meta.url).pathname;
```
with
```js
 const pathname = new URL(req.originalUrl, server.resolvedUrls?.local.toString())
```

### DIFF
--- a/demo/package.json
+++ b/demo/package.json
@@ -65,7 +65,7 @@
     "@codingame/monaco-vscode-gulp-default-extension": "file:../dist/default-extension-gulp",
     "@codingame/monaco-vscode-handlebars-default-extension": "file:../dist/default-extension-handlebars",
     "@codingame/monaco-vscode-hlsl-default-extension": "file:../dist/default-extension-hlsl",
-    "@codingame/monaco-vscode-html-default-extension": "file:../dist/default-extension-html",
+        "@codingame/monaco-vscode-html-default-extension": "file:../dist/default-extension-html",
     "@codingame/monaco-vscode-html-language-features-default-extension": "file:../dist/default-extension-html-language-features",
     "@codingame/monaco-vscode-ini-default-extension": "file:../dist/default-extension-ini",
     "@codingame/monaco-vscode-ipynb-default-extension": "file:../dist/default-extension-ipynb",
@@ -181,6 +181,7 @@
     "ansi-colors": "^4.1.3",
     "dockerode": "^4.0.2",
     "express": "^4.18.2",
+    "import-meta-resolve": "^4.0.0",
     "monaco-editor": "^0.45.0",
     "vscode": "file:../dist/main",
     "ws": "^8.16.0"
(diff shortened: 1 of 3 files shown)


demo/vite.config.ts
```diff
@@ -1,110 +1,111 @@
...
+      allow: ['../'] // allow to load codicon.ttf from monaco-editor in the parent folder
+    }
+  },
+  define: {
+    rootDirectory: JSON.stringify(__dirname)
+  },
+  resolve: {
+    dedupe: ['monaco-editor', 'vscode', ...localDependencies]
+  }
+})
```
CGNonofr, line 111: Why is github not able to display a proper diff on this file?

dkattan: I may have messed up the line endings. I can fix it now, give me a few minutes. I'll rebase while I'm at it.

CGNonofr: Did you forgot to push? :)

CGNonofr: It still doesn't seem to be ok :thinking:

demo/vite.config.ts
```diff
@@ -1,110 +1,111 @@
...
+            const code = fs.readFileSync(args.path, 'utf8')
+
+            const assetImportMetaUrlRE = /\bnew\s+URL\s*\(\s*('[^']+'|"[^"]+"|`[^`]+`)\s*,\s*import\.meta\.url\s*(?:,\s*)?\)/g
+            let i = 0
+            let newCode = ''
+            for (let match = assetImportMetaUrlRE.exec(code); match != null; match = assetImportMetaUrlRE.exec(code)) {
+              newCode += code.slice(i, match.index)
+
+              const path = match[1].slice(1, -1)
+              const resolved = await resolve(path, url.pathToFileURL(args.path).toString())
```
CGNonofr, line 83: it seems the resolve function is not async

demo/package.json
```diff
@@ -181,6 +181,7 @@
     "ansi-colors": "^4.1.3",
     "dockerode": "^4.0.2",
     "express": "^4.18.2",
+    "import-meta-resolve": "^4.0.0",
```
CGNonofr, line 184: it should probably be a dev dependency

CGNonofr (commented):
Seems to still build properly on Linux :+1:

CGNonofr:
This plugin was externalized and used in https://github.com/CodinGame/monaco-vscode-api/pull/329

PR #318, opened 2024-01-17
Bundle node build of default extensions with ext-host server

CGNonofr:
@kaisalmen also use the full VSCode building process for web extension, which:
- removes extensions not usable in the browser at all (11)
- fix the sourcemap error it seems (but just because it's an https url instead of a path but still)

As the extension codes are now minified, it's harder to detect which files are used or not by parsing the code, so I've modified the rollup plugin to include any existing files (it's simpler and more robust that way)

It also rollbacks to node 18 (the same used by VScode), the 21 has some issues with the ext-host server due to a `navigator` empty global variable existing, preventing VSCode from detection the OS, and the v20 produce build issues (really hard to investigate)

And finally, it builds and packages defaults extensions with the ext-host server. I've added an option to prevent it from loading them all at startup. There is still a way to load those extension one by one using the `registerRemoteExtension` function


kaisalmen:
@CGNonofr during `npm i` I noticed this:
image

Does the `--no-yarn` option work.

CGNonofr:
> @CGNonofr during `npm i` I noticed this: image
> 
> Does the `--no-yarn` option work? It seems to work, though, but pollutes the output.

The warning comes from the VSCode build process (@vscode/vsce not being up to date, it was fixed in https://github.com/microsoft/vscode-vsce/pull/807). It happens when building VSCode as well

CompuIves:
Wow, perfect timing, I was just looking into this! I can review this tomorrow.

CompuIves (approved):
Code looks good, also tested the demo and everything works as expected (including the built-in server extensions). Thanks for introducing this change!

PR #327, opened 2024-01-24
Include monaco-editor, remove it from dependencies

CGNonofr:
Related to https://github.com/CodinGame/monaco-vscode-api/discussions/325

@kaisalmen here's the POC

I'd like to know what you think about it

and here's a draft of what I plan to put in the release description:

***

## Description

Remove dependency to `monaco-editor`: `monaco-editor` is now bundled by the lib.

A `@codingame/monaco-vscode-editor-api` package now exists to access the `monaco-editor` api

## Migration guide:

- replace `monaco-editor` by the provided api package:
```bash
npm install monaco-editor@npm:@codingame/monaco-vscode-editor-api
```
- `monaco.editor.create` can now be used instead of `createConfiguredEditor`

***


tsconfig.types.json
```diff
@@ -47,15 +47,12 @@
...
-		],
-		"monaco-editor/esm/*": [
-			"./vscode/src/*"
 		]
 	},
 	"declaration": true,
 	"emitDeclarationOnly": true,
 	"skipLibCheck": true,
-	"types": ["monaco-editor", "node"]
+	"types": ["node"]
```
kaisalmen, line 55: I learnt this quite recently. Are you aware that this prevents the usage of any types defined in `node_modules/types`. It overrides it hard (see: https://www.typescriptlang.org/tsconfig#types). If you do this on purpose nevermind my comment.

src/override/vs/base/common/marked/marked.ts
```diff
@@ -0,0 +1,4 @@
+import marked from 'vscode-marked'
```
kaisalmen, line 1: Can you explain what this does / what "feat: replace inlined marked by dependency" really means? Thanks.

package.json
```diff
@@ -53,19 +51,18 @@
     "@rollup/plugin-node-resolve": "^15.2.3",
     "@rollup/plugin-replace": "^5.0.5",
     "@rollup/plugin-typescript": "^11.1.6",
-    "@types/diff": "^5.0.9",
     "@types/mime-types": "^2.1.4",
     "@types/node": "^18.17.15",
     "@types/semver": "^7.5.6",
-    "@types/vscode-semver": "npm:@types/semver@=5.5.0",
     "@types/vscode": "~1.84.2",
```
kaisalmen, line 57: Why not `1.85.0`?

kaisalmen (commented):
Great work! :rocket: I will test the preview version with `monaco-languageclient` and `monaco-editor-wrapper` and let you know of any issues I may encounter.

We should consider a couple of things:
- We should gather feedback from users of `monaco-vscode-api`. @CompuIves your feedback will be very valuable here, I think.
- Should we ask users to extensively test the preview version before making the official release?
- Users of `@codingame/monaco-vscode-api` and `monaco-languageclient` should have been already aware they got/needed an altered monaco-editor and therefore should be just fine with the change. Are we overlooking something big here or can we just state: If you want to use this library stack that is the way it is?

src/override/vs/base/common/marked/marked.ts
CGNonofr: Sorry if it not clear.

What was done here is very similar to what is already done for semver. For some reasons, VSCode inlines marked and semver in their codebase. It's an issue for us because the inlined version is in CommonJS and the rollup treeshaking breaks it (because of the use of IIFEs).

Any file in the `override` directory will be used instead of the corresponding VSCode file, I use it to use the official semver package (with the corresponding old version) instead of the file inlined in VSCode

I don't find any documentation on why they do it, [I've ask but I don't really expect any answer soon](https://stackoverflow.com/questions/77872402/why-is-semver-and-marked-inlined-in-vscode)

tsconfig.types.json
CGNonofr: I... don't know, I don't think it was done on purpose.. I'll remove it after I check it causes no trouble

package.json
CGNonofr: The types weren't available at the time the update was make, and it's excluded for ncu so I didn't noticed it, will update

CGNonofr: 5ada4c15afcb1d475adc51a202647290c0a4ce81

tsconfig.types.json
kaisalmen: FYI/warning; This can have effects/compilation problems as suddenly more types are visible for `tsc`

src/override/vs/base/common/marked/marked.ts
kaisalmen: Thank you. Did you ask that at stackoverflow because there was no answer in the VSCode repo?

tsconfig.types.json
CGNonofr: I've check the build of the main project and it's fine, but I should have check the build of the demo, hold on...

src/override/vs/base/common/marked/marked.ts
CGNonofr: I've ask to stackoverflow because the issue workflow on the VSCode repo redirects to it when you want to ask questions

kaisalmen: Thanks, I didn't know that

tsconfig.types.json
CGNonofr: 1a005f79c83032db5b628922640649fa3526a1cc

kaisalmen: 👍

CGNonofr:
> Great work! 🚀 I will test the preview version with `monaco-languageclient` and `monaco-editor-wrapper` and let you know of any issues I may encounter.

Thank you! :)

> We should consider a couple of things:
> 
> * We should gather feedback from users of `monaco-vscode-api`. @CompuIves your feedback will be very valuable here, I think.

Totally agree!

> * Should we ask users to extensively test the preview version before making the official release?

I don't know, it will release a major version so... anyone having issues with it can just not upgrade for the moment

> * Users of `@codingame/monaco-vscode-api` and `monaco-languageclient` should have been already aware they got/needed an altered monaco-editor and therefore should be just fine with the change. Are we overlooking something big here or can we just state: If you want to use this library stack that is the way it is!?

I don't think it complexifies the usage and I'm pretty convinced this is 99% backward compatible (1% being the internal monaco-editor module import -  and we will able to fix it later module per module). I hope we don't overlook anything but I really don't know what it could be that we won't be able to fix

kaisalmen:
@CGNonofr encountered a fist issue:
main `vscode` package does not export the editor in its package. It works in the demo, but not with the npm package:

image

CGNonofr:
> @CGNonofr encountered a first issue: main `vscode` package does not export the editor in its `package.json`. It works in the demo, but not with the npm package:
> 
> image

Ok thanks, there was indeed an issue, should be fixed by f95b460fd1f5ead7a54704f65a22a6758e5e748c, testing...

CGNonofr:
Ok @kaisalmen it seems we SHOULDN'T rewrite the history of a PR or semantic release will fail to release another next version from the branch... (and I didn't find a way to recover it)

So I've currently publishing from another branch: v2 so the deployed version will be 2.0.0-v2.1

Can you give it a try?

kaisalmen:
> Can you give it a try?

It works, see https://github.com/TypeFox/monaco-languageclient/pull/599

CGNonofr:
> > Can you give it a try?
> 
> It works, see [TypeFox/monaco-languageclient#599](https://github.com/TypeFox/monaco-languageclient/pull/599)

Ok, so what's the next step?

kaisalmen:
> Ok, so what's the next step?

We should merge this PR and https://github.com/TypeFox/monaco-languageclient/pull/599 . I am currently moving the wrapper end merging the example into the monaco-languageclient While doing this I will see if there are further issues. But, I don't know if I can make it this week.

We should either let other people test the pre-release, gather feedback and then release the final version at some point OR you can also release the final version soon as it is **really a distinguishable major version** and we report and solve potential issues later and release updates. What do you prefer?

kaisalmen (approved):
LGTM. It works as expected.

kaisalmen:
> I think I prefer to move fast, break things (big changes will eventually have an end)

With regard to `monaco-languageclient` this major version will only become relevant once version `8.0.0` is released and it will likely not be this week. More likely end of next week.

I guess you / Codingame is the major user, so will immediately notice something is broken. Others should be aware that switching to a new major version may break things.

CGNonofr:
> > I think I prefer to move fast, break things (big changes will eventually have an end)
> 
> With regard to `monaco-languageclient` this major version will only become relevant once version `8.0.0` is released and it will likely not be this week. More likely end of next week.
> 
> I guess you / Codingame is the major user, so will immediately notice something is broken. Others should be aware that switching to a new major version may break things.

Exactly!

But there is so many ways to use the lib and so many features that it's really hard to guarantee no regression...

Also I think we're reactive enough on bug reports

kaisalmen:
> Also I think we're reactive enough on bug reports

💯

PR #328, opened 2024-01-25
Only bundle used files when there is no extension script

CGNonofr:
Recently, I've made a change to bundle all extension files in Rollup plugins.

I works well except if you try to load an extension designed for vscode desktop containing very big file (like the clang extension containing a >100mo binary).

You may want to load such extension to get the languages and grammars but don't care about the extension javascript code.

For instance, the `Angular Language Service` contains grammars extension for inline template but also the whole language server.

What this PR do is basically restoring the previous behavior when there is no javascript code designed to run in the browser inside the extension. In that case, it's possible to know all resources used by the manifest contributes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,6 +27,7 @@
         "@xterm/headless": "5.4.0-beta.14",
         "@xterm/xterm": "5.4.0-beta.14",
         "cookie": "^0.5.0",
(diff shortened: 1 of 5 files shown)


kaisalmen (approved):
Let's get this in first? I will start reviewing the other PRs now.

PR #329, opened 2024-01-26
Use published esbuild plugin

### LINKED ISSUE #179: Expose vite dev server plugin?
@CGNonofr Is it possible to expose the vite dev server `import.meta.url` [workaround plugin](https://github.com/CodinGame/monaco-vscode-api/blob/main/demo/vite.config.ts#L63-L88) as another npm sub export of `monaco-vscode-api`?

I have copied an now outdated version and used it at different places. I also just realized that when people used vscode default-extension in a vite dev server they all have the same problems. It is easier to tell them, please use this sub export from `monaco-vscode-api` instead of copying code that gets outdated. WDYT?

CGNonofr:
fix #179

### DIFF
--- a/demo/package-lock.json
+++ b/demo/package-lock.json
@@ -158,6 +158,7 @@
         "ws": "^8.16.0"
       },
       "devDependencies": {
+        "@codingame/esbuild-import-meta-url-plugin": "^1.0.2",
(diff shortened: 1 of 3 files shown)


kaisalmen (approved):
Wonderful! LGTM 🎉

PR #336, opened 2024-02-02
Improve modularity

CGNonofr:
This PR removes the `include` and `exclude` query param when importing VSCode module that allowed us to import only partially a module. It has some downside:
- It's dangerous: it you import with `include=anyFunction` and you import it without any filter elsewhere, the code matching `anyFunction` will be duplicated
- The filter is really basic (only on called function name)
- If the function name change during a VSCode update (it will happen in the next VSCode release), this mechanism break silently

Replaced by simply splitting the file by hands using the patch file

This PR aims at removing unwanted visible commands (in the command picker) when the view service override is not used

There are still some remainings but I don't think will be able to perfectly split them without introducing a huge complexity


CompuIves (approved):
This PR makes sense. There's also a downside to using the patch file, though. We need to keep the new files in sync, and make sure that when we update we check that those files are still in sync with VSCode equivalent.

That said, I think those separate files is the lesser of the two evils, it's better than having `?include`. I was bitten before by that as well.

CGNonofr:
> This PR makes sense. There's also a downside to using the patch file, though. We need to keep the new files in sync, and make sure that when we update we check that those files are still in sync with VSCode equivalent.
> 
> That said, I think those separate files is the lesser of the two evils, it's better than having `?include`. I was bitten before by that as well.

I am looking for a better approach than having a big PATCH file... relying on a fork with several additional commit and rebasing it at every VSCode release seems like a better idea but I don't know how to implement it exactly

PR #350, opened 2024-02-16
Display extensions as builtin

CGNonofr:
Display extensions registered with `registerExtension` as builtin in the extension panel:

image


CompuIves (approved):
Clean PR, makes sense to me! I'm working on something related to this now, I'm exploring if I can _just_ use the extension host server (instead of the whole vscode server). I might open an issue/PR related to it.

CGNonofr:
> I'm exploring if I can _just_ use the extension host server (instead of the whole vscode server). I might open an issue/PR related to it.

You mean on the server side or on the client side? can't wait to see it!

PR #357, opened 2024-02-27
Fix standalone diff editor style

CGNonofr:


kaisalmen (approved):
Thank you. LGTM. I will report if there are further issues.

kaisalmen:
@CGNonofr I confirm that this works fine with regular and diff editor.

PR #370, opened 2024-03-14
Vscode 1.87.2

CGNonofr:
switch to netlify to be able to set headers allowing the typescript worker to use shared array buffer and enable project wide intellisense


kaisalmen (approved):
Too bad GH Pages does support setting headers. Tested new deployment. LGTM

Issue #371, opened 2024-03-16
Default search service extension - follow file link doesn't work


Abdallatif:
When using the `@codingame/monaco-vscode-search-result-default-extension` and open search results in editor, then hover over a file name and click on the follow link

![image](https://github.com/CodinGame/monaco-vscode-api/assets/10572734/b28385eb-9bed-416d-b755-599aebf2624e)

The following error appears:

![image](https://github.com/CodinGame/monaco-vscode-api/assets/10572734/7148ee01-e485-4377-ac1a-0e7eda1ca6b4)

In the console, it shows the following error:
```
Uncaught (in promise) Error: Unable to read file '/test.js' (Error: Unable to resolve nonexistent file '/test.js')
    at VHe.restoreReadError (main.common-7DLNmK_d.js:1427:23935)
    at VHe.doReadFileStream (main.common-7DLNmK_d.js:1427:23642)
    at async WHe.doRead (main.common-7DLNmK_d.js:1427:78870)
    at async WHe.readStream (main.common-7DLNmK_d.js:1427:78625)
    at async ame.resolveFromFile (main.common-7DLNmK_d.js:1427:57239)
    at async ame.resolve (main.common-7DLNmK_d.js:1427:55401)
    at async FHe.doResolve (main.common-7DLNmK_d.js:1427:73374)
    at async sKe.doCreateReferencedObject (main.common-7DLNmK_d.js:1734:6398)
    at async Xjt.acquire (main.common-7DLNmK_d.js:29:3253)
    at async oKe.createModelReference (main.common-7DLNmK_d.js:1734:8629)
```

I tried to do some debugging and find the core issue, my guess here is that the `~` is being ignored while it should be converted to the workspace uri.

I also found (by manually changing the editor search results) that if we remove the `~/` from the file path, vscode would be able to locate the file correctly.

CGNonofr:
Thanks for the report!

Once again, it's a bug in VSCode (or a combination of bugs) that doesn't affect VScode in the way it's used but does impact monaco-vscode-api

Explanation:
- in web mode, VSCode try to guess the local user home based on what it has access to. It there is a remote connection, it takes the remote user home but if not, it take the first workspace folder as user home, or the folder containing the workspace configuration
[...]

PR #373, opened 2024-03-19
Add a way to configure user home

CGNonofr:
workaround for #371


kaisalmen (approved):
Reset layout button works nicely. Changes look good to me. I didn't test to configure user home, though. LGTM

PR #375, opened 2024-03-21
Fix editor api types

CGNonofr:
Fix for the demo that was broken in the last MR as well as properly checking demo build in PRs


kaisalmen:
@CGNonofr just updated deps in `monaco-languageclient` and saw this. Is this the same issue?
image

CGNonofr:
> @CGNonofr just updated deps in `monaco-languageclient` and saw this. Is this the same issue? image

Not at all, that's something else :eye:

kaisalmen:
Ok, if I add the `xterm....` dependencies as devDependencies all but the last go away. I will clear the node_modules folder and `package-lock.json` to exclude I am Don Quijote and chase windmills

kaisalmen:
> I will clear the node_modules folder and package-lock.json to exclude I am Don Quijote and chase windmill

It is not thath

CGNonofr:
@kaisalmen 

here's the fix for the last error: 30f05b88b914368757c2ff8dd4e0e7902f9aec68

regarding the other errors, there is indeed a file in monaco-vscode-api that references xterm without having it as dependency, but that file is not supposed to be imported by the package itself. Only by terminal and testing service override that have xterm as dependency. But I guess it can happens if you're making typescript check every files in node_modules, even those unused.

kaisalmen:
I don't not change the compiler config. Only types for `vscode` are considered. That is the showconfig output for mlc:
image

CGNonofr:
Having a look!

kaisalmen (approved):
Is the xterm problem fixed as well now?

CGNonofr:
> Is the xterm problem fixed as well now?

Not really, the issue is that the module importing it it references by vscode/services to be able to export the service interface... maybe it's not a so good idea to export all service interfaces from there... what do you think?

kaisalmen:
> maybe it's not a so good idea to export all service interfaces from there... what do you think?

Did you change that after `3.1.2`? I just looked at the delta and don't see why or is it happening in the rollup magic?

CGNonofr:
> > maybe it's not a so good idea to export all service interfaces from there... what do you think?
> 
> Did you change that after `3.1.2`? I just looked at the delta and don't see why behavior/visibility should have changed or is it happening in the rollup magic?

Not at all, It's not checked by the demo here because those dependencies are installed, maybe the build process changed on your side?

0dd994c7a53ec990105b692fbe390242eeed4c8d should help as a workaround until we decide what to do with service interfaces

kaisalmen (approved):
Let's try this. The awkward thing is that even with `types: []` the problem appears. No changes in the build process, see: https://github.com/TypeFox/monaco-languageclient/tree/mva-32x

CGNonofr:
> Let's try this. The awkward thing is that even with `types: []` the problem appears. No changes in the build process, see: https://github.com/TypeFox/monaco-languageclient/tree/mva-32x

it's a mystery :shrug: I'm not surprised setting types to empty array does nothing as the problematic file is referenced transitively in the code, there is no way to ignore it (except `skipLibCheck` ofc)

kaisalmen:
> there is no way to ignore it (except skipLibCheck ofc)

very true, but I don't want that and it just hides the problem here and it will pop-up elsewhere

kaisalmen:
@CGNonofr I confirm it builds again.

PR #378, opened 2024-03-25
Update monaco-editor-webpack-plugin descripton with regard to languages

kaisalmen:
This updates the `languages` bullet in the monaco-editor-webpack-plugin section. I also applied some markdown formatting recommendation.
Do you want to mention the monaco-languages somewhere else/some other section?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,10 @@ Make sure `typeof process === 'undefined'` returns true when using this library.
 Starting from v2, [monaco-editor-webpack-plugin](https://www.npmjs.com/package/monaco-editor-webpack-plugin) can't be used
 
 Here's the alternative for each options:
+
 - `filename`: it can be configured at the webpack level directly
 - `publicPath`: it can be configured at the webpack level or by hands when redistering the worker in `window.MonacoEnvironment`.
-- `languages`: Monaco languages cannot be used anyway. Import vscode language extensions instead (`@codingame/monaco-vscode-xxx-default-extension`).
+- `languages`: Import vscode language extensions (`@codingame/monaco-vscode-xxx-default-extension`) or (`@codingame/@codingame/monaco-vscode-standalone-*`), but please obey that monaco languages can only be used if `themes` and `textmate` service overrides are not configured.
 - `features`: With this lib, you can't remove editor features.
 - `globalAPI`: you can set `window.MonacoEnvironment.globalAPI` to true
 
@@ -40,12 +41,14 @@ Here's the alternative for each options:
 This library uses a lot the `new URL('asset.extension', import.meta.url)` syntax which [is supported by vite](https://vitejs.dev/guide/assets.html#new-url-url-import-meta-url)
 
 While it works great in `build` mode (because rollup is used), there is some issues in `watch` mode:
+
 - import.meta.url is not replaced while creating bundles, it is an issue when the syntax is used inside a dependency
 - vite is still trying to inject/transform javascript assets files, breaking the code by injecting ESM imports in commonjs files
 
 There are workarounds for both:
 
 - We can help vite by replacing `import.meta.url` by the original module path:
+
 ```typescript
 import importMetaUrlPlugin from '@codingame/esbuild-import-meta-url-plugin'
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,9 +29,10 @@ Make sure `typeof process === 'undefined'` returns true when using this library.
 Starting from v2, [monaco-editor-webpack-plugin](https://www.npmjs.com/package/monaco-editor-webpack-plugin) can't be used
 
 Here's the alternative for each options:
+
 - `filename`: it can be configured at the webpack level directly
 - `publicPath`: it can be configured at the webpack level or by hands when redistering the worker in `window.MonacoEnvironment`.
-- `languages`: Monaco languages cannot be used anyway. Import vscode language extensions instead (`@codingame/monaco-vscode-xxx-default-extension`).
+- `languages`: Import vscode language extensions (`@codingame/monaco-vscode-xxx-default-extension`) or (`@codingame/@codingame/monaco-vscode-standalone-*`), but please obey that monaco languages can only be used if `themes` and `textmate` service overrides are not configured.
```
CGNonofr, line 35: And the opposite as well though: language VSCode extensions can't be used without those service overrides

README.md
```diff
@@ -77,16 +81,17 @@ See [this issue](https://github.com/CodinGame/monaco-vscode-api/issues/186) or t
...
 
 At least thoses files should have the headers:
+
 - the main page html
 - the worker extension host iframe html: `webWorkerExtensionHostIframe.html`
 - the worker extension host worker javascript: `extensionHost.worker.js`
 
-If adding those headers is not an options, you can have a look at https://github.com/gzuidhof/coi-serviceworker
-
+If adding those headers is not an options, you can have a look at <https://github.com/gzuidhof/coi-serviceworker>
```
CGNonofr, line 94: I've realized it's not a good idea as it conflicts with the VSCode webview service worker :-\ maybe it's worth mentionning it while you're at it

CGNonofr (commented):
Thank you!

> Do you want to mention the monaco-languages somewhere else/some other section?

I was not sure... that's not something I would like to recommend since it locks a lot of features... :-\ I'm also afraid that it would complexity even more the README and make it more confusing. Maybe if an `advanced` section or something? I don't know

The README probably needs a deep rework, that's not really something easy for me

kaisalmen:
@CGNonofr Does it make sense to exchange build and lint? My commit message was not well-formed

README.md
kaisalmen: Yes, but that is already stated in the corresponding section. Should I refine this here, too?

CGNonofr:
> @CGNonofr Does it make sense to exchange build and lint? My commit message was not well-formed

You mean to change the order? what would it change? aren't they done in // ?

README.md
kaisalmen: Do you just want to drop the sentence? You removed coi-serviceworker again completely here, didn't you?

README.md
CGNonofr: If there is a way to say it in a not confusing way, maybe? some redundancy can't hurt in the doc :)

CGNonofr:
> > what would it change?
> 
> Build takes 16 min. Lint only some seconds. It would fail fast. That's all. 🙂

I'm not sure to understand, you are talking about commitlint, right?

README.md
kaisalmen: I can do confusing for sure, but I will try to make it not confusing. 🤣

README.md
CGNonofr: I've removed it because it conflicts with webviews, but not everyone is using webviews

kaisalmen:
> I'm not sure to understand, you are talking about commitlint, right?

It's Monday. Nevermind, It is already where it should be. I just checked the actions instructions-

README.md
kaisalmen: What about:

"If adding those headers is not an options, you can have a look at <https://github.com/gzuidhof/coi-serviceworker>, but only if you are not using webviews as it introduces problems then."

CGNonofr: Yep, seems good and not confusing (:

kaisalmen: 👍

kaisalmen: Done ⬇️

README.md
kaisalmen: Done ⬇️

CGNonofr (approved):

CGNonofr:
Thanks!

PR #380, opened 2024-03-27
Use official ext host server

CGNonofr:
There is no point of building the server when Microsoft already does it for us

It allows to simplifies the build process but also simplifies the usage in a docker environment: you don't need node or npm, the VSCode server is a standalone


CompuIves:
Oh, this will simplify the build process for us as well!

CGNonofr:
> Oh, this will simplify the build process for us as well!

Can you elaborate?

CompuIves (approved):
This PR speaks to how far you've come with reintegrating VSCode in a modular way. That it can interface with an official VSCode server is really cool. Changes look good to me!

CompuIves:
> Can you elaborate?

I'm working on VSCode Server support in CodeSandbox, and we have a build step where we build vscode server and tar it, so the server can download it. We need to build VSCode server with an old version of Node, though, because our current environment has a glibc version that's too high. Also, we need to ship standalone Node to our VMs. So this would remove both steps (building in Docker, and shipping standalone Node ourselves).

kaisalmen:
@CGNonofr nice work. Was mostly AFK over extended easter weekend.

PR #382, opened 2024-04-05
feat: add secret storage service

CompuIves:
This adds the `ISecretStorage` service override. In the browser this is just plain storage in memory. This service is needed by some extensions (like the Thunder Client) extension.

An implementation can actually be passed when initializing VSCode, if that's not available it will default to storing the secrets in-memory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,8 @@ Additionally, several packages that include the VSCode version of some services 
   - Update detection, release notes...
 - **Localization**
   - Register callbacks to update the display language from the VSCode UI (either from the `Set Display Language` command or from the extension gallery extension packs)
+- **Secret Storage**
+  - Storage of secrets for extensions, will store by default in-memory. You can pass a custom implementation as part of the workbench construction options when initializing monaco services (under `secretStorageProvider`).
 
 Usage:
 
--- a/src/service-override/secretStorage.ts
+++ b/src/service-override/secretStorage.ts
@@ -1,0 +1,15 @@
+import { IEditorOverrideServices } from 'vs/editor/standalone/browser/standaloneServices'
+import { IEncryptionService } from 'vs/platform/encryption/common/encryptionService'
+import { SyncDescriptor } from 'vs/platform/instantiation/common/descriptors'
+import { ISecretStorageService } from 'vs/platform/secrets/common/secrets'
+import { EncryptionService } from 'vs/workbench/services/encryption/browser/encryptionService'
+import { BrowserSecretStorageService } from 'vs/workbench/services/secrets/browser/secretStorageService'
+
+export { LocalStorageSecretStorageProvider } from 'vs/code/browser/workbench/workbench'
+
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -236,6 +236,8 @@ Additionally, several packages that include the VSCode version of some services
   - Survey/feedback support
 - **Update**
   - Update detection, release notes...
+- **Secret Storage**
+  - Storage of secrets for extensions, will store by default in-memory, unless passed a custom implementation in the environment service.
```
CGNonofr, line 240: `unless passed a custom implementation in the environment service` is probably a little bit confusing, are you talking about the WorkbenchContructionOptions given to the `initialize` function?

src/service-override/secretStorage.ts
```diff
@@ -0,0 +1,13 @@
...
+import { ISecretStorageService } from 'vs/platform/secrets/common/secrets'
+import { EncryptionService } from 'vs/workbench/services/encryption/browser/encryptionService'
+import { BrowserSecretStorageService } from 'vs/workbench/services/secrets/browser/secretStorageService'
+
+export default function getServiceOverride (): IEditorOverrideServices {
+  return {
+    [ISecretStorageService.toString()]: new SyncDescriptor(BrowserSecretStorageService, [], true),
+    [IEncryptionService.toString()]: new SyncDescriptor(EncryptionService, [], true)
+  }
+}
```
CGNonofr, line 13: What about re-exporting the `LocalStorageSecretStorageProvider` from VSCode?

CGNonofr (changes requested):
Thank you for your contribution! Just a few minor comments

src/service-override/secretStorage.ts
CompuIves: Good idea! Will do that.

README.md
CompuIves: Yep, I'll make this more clear.

src/service-override/secretStorage.ts
CompuIves: Okay, hmm, so it looks like the `LocalStorageSecretStorageProvider` is not part of the https://github.com/microsoft/vscode/blob/main/src/vs/code/browser/workbench/workbench.ts#L168C14-L168C47). I can't seem to import it here, because it's not included in the VSCode sources we install. I _can_ copy over the code and use that though, what do you think?

CGNonofr: we're just removing the whole vs/code directory in the install-vscode script, we can probably make it more specific

CompuIves:
Applied the feedback!

CGNonofr (approved):

src/service-override/secretStorage.ts
CGNonofr: It wasn't a good idea, because it makes us pull the whole workbench... and it seems that provider rely on a dom element existing anyway, so It's maybe too specific

PR #385, opened 2024-04-09
Localization service override

CGNonofr:


kaisalmen (approved):
Nice, changing the language works as expected. I need to adjust the mlc examples, but this change is very nice. Thank you. LGTM.

PR #386, opened 2024-04-09
Fix localization issues

CGNonofr:


src/service-override/localization.ts
```diff
@@ -2,12 +2,12 @@ import { IEditorOverrideServices } from 'vs/editor/standalone/browser/standalone
 import { SyncDescriptor } from 'vs/platform/instantiation/common/descriptors'
 import { ILanguagePackItem, ILanguagePackService } from 'vs/platform/languagePacks/common/languagePacks'
 import { ILocaleService } from 'vs/workbench/services/localization/common/locale'
-import { URI } from 'vs/workbench/workbench.web.main'
```
kaisalmen, line 5: > Ok I've found an issue with the new service override!

Yeah that is likely it.

kaisalmen (approved):
Thank you. LGTM.

src/service-override/localization.ts
CGNonofr: It is ! :)

PR #387, opened 2024-04-09
Fix localization issues second attempt

CGNonofr:


kaisalmen:
@CGNonofr awesome, when do you sleep? I have one further request: Can you export interfaces `AvailableLanguage` and `LocalizationOptions` from `@codingame/monaco-vscode-localization-service-override` because this is part of the exposed `getServiceOverride`, but you can't import the type. I externalized the "default" function and it is nicer if it can have a specific type. Thank you.

CGNonofr:
> @CGNonofr awesome, when do you sleep? I have one further request: Can you export interfaces `AvailableLanguage` and `LocalizationOptions` from `@codingame/monaco-vscode-localization-service-override` because this is part of the exposed `getServiceOverride`, but one can't import the type. I externalized the "default" function and it is nicer if it can have a specific type. Thank you.

sure! 87a8fa6e987ae8a01240a01ff20d21f7e4434866

kaisalmen (approved):
Thank you once more. LGTM

PR #396, opened 2024-04-15
Dedup non-exclusive external dependencies

CGNonofr:
vscode-marked and vscode-semver are now duplicated in service-overrides that depend on it


kaisalmen:
@CGNonofr the demo build broke.

CGNonofr:
> @CGNonofr the demo build broke.

Oopsy, should be fine now :crossed_fingers:

PR #409, opened 2024-04-17
Improve filesystem management

### LINKED ISSUE #406: Improve filesystem management
- Add documentation on how the overlay works
- Allow to replace the overlay FS by something else
- Allow to remove the default memory overlay

CGNonofr:
fix #406

It allows to implement the filesystem exactly the way you want

It also allows to fully use the HTMLFileSystemProvider, so that going to `?htmlFileSystemProvider=true&mode=full-workbench` should be REALLY close to vscode.dev


kaisalmen (approved):
Looks good. Nice work. I tested it locally and it does exactly what it should. LGTM 🎉

PR #412, opened 2024-04-18
Make parameter optional

CGNonofr:
My bad :-\


CGNonofr:
> Wasn't really wrong, but let's go.

It just wasn't backward compatible

kaisalmen:
> It just wasn't backward compatible

Yes, consistent with semantic versioning, but better to be bw-comp. 👍

PR #413, opened 2024-04-19
Fix encoding management

CGNonofr:
- Restore the usage of @vscode/iconv-lite-umd/jschardet to detect/handle different encoding (they are lazy-loaded)
- Remove some mocked file
- Attempt to move as much code as possible to the leaf packages: it allowed to remove almost 2mo from the main package


kaisalmen (changes requested):
Local test look good and potentially smaller bundles are very welcome by end-users.
You introduced a new service and did not document it that's why I request changes.

CGNonofr:
> Local test look good and potentially smaller bundles are very welcome by end-users. You introduced a new service and did not document it that's why I request changes.

Since it's an internal service override, I was wondering if it was worth it to document it 🤔

kaisalmen:
> Since it's an internal service override, I was wondering if it was worth it to document it 🤔

Yes, that makes sense

kaisalmen (approved):
LGTM

kaisalmen:
@CGNonofr Are more changes upcoming or is there another reason you did not merge this, yet?

CGNonofr:
> @CGNonofr Are more changes upcoming or is there another reason you did not merge this, yet?

There are some... experiments, but if it leads to anything it'll be in another MR

kaisalmen:
> We can release anytime if you want! I don't have anything else in the pipe though :)

We can combine it with other PR. No pressure from my side. I am busy with something else this week.

PR #417, opened 2024-04-30
Optimize localization

CGNonofr:
Apply the same optimization than VScode in vscode.dev, which replaces string translation keys by their indexes within the module

in the localization file:
```
{
  "vs/common/random/module": {
    "some.random.translation.key": "The actual wording",
    "some.other.translation.key": "The other wording"
  }
}
```
is replaced by
```
{
  "vs/common/random/module": [
    "The actual wording",
    "The other wording"
  ]
}
```

and in the modules:
```
localizeWithPath(moduleId,  "some.random.translation.key", "The default wording")
```
is replaced by
```
localizeWithPath(moduleId,  0,  "The default wording")
```

This PR:
- reduce the size of the main package alone by about 400ko (not counting the service override packages)
- reduce each localization package by about 300ko

Note that the localization keys can't be minified and neither be compressed very well (each key is different) so the final compressed size ratio should be even bigger


kaisalmen (approved):
Thank you. Tested locally. Have not seen anything obvious. Changing languages works nicely. I did not perform an in-depth review of the code. Will test the new release as usual.

PR #428, opened 2024-05-07
Prevent infinite loop on readdir

### LINKED ISSUE #424: Deployed demo stalls browser
I have seen that the [deployed demo](https://monaco-vscode-api.netlify.app/) stalls Chrome and Firefox on Windows. Console logs are inconclusive. Some loading take suddenly very long ⬇️ and then interaction with the page is then not possible.
This does not happen with Firefox and Chrome on Linux. 🤷‍♂️ 
image

CGNonofr:
fix #424

### DIFF
--- a/src/service-override/files.ts
+++ b/src/service-override/files.ts
@@ -194,8 +194,9 @@ class RegisteredFileSystemProvider extends Disposable implements IFileSystemProv
 
   public async readdir (resource: URI): Promise<[string, FileType][]> {
     const includedPaths = Array.from(this.files.keys())
-      .map(uri => extUri.relativePath(resource, URI.parse(uri))!)
-      .filter(path => !path.startsWith('..'))
+      .map(uri => URI.parse(uri))
+      .filter(uri => uri.path.startsWith(resource.path))
+      .map(uri => extUri.relativePath(resource, uri)!)
 
     const files = includedPaths.filter(path => !path.includes('/'))
     const directories = includedPaths.filter(path => path.includes('/')).map(path => path.slice(0, path.indexOf('/')))


src/service-override/files.ts
```diff
@@ -194,8 +194,9 @@ class RegisteredFileSystemProvider extends Disposable implements IFileSystemProv
 
   public async readdir (resource: URI): Promise<[string, FileType][]> {
     const includedPaths = Array.from(this.files.keys())
-      .map(uri => extUri.relativePath(resource, URI.parse(uri))!)
-      .filter(path => !path.startsWith('..'))
+      .map(uri => URI.parse(uri))
+      .filter(uri => uri.path.startsWith(resource.path))
+      .map(uri => extUri.relativePath(resource, uri)!)
```
CGNonofr, line 199: For some reasons, `relativePath` returns the absolute path if the second parameter is not a child of the first parameter, and not a path starting with `..`

kaisalmen: Strange indeed. How did you find that problem in the first place?

kaisalmen (approved):
It is working now. WSL serving the demo and working with Ubuntu Chrome and Windows Chrome. Wasn't doing this yesterday. Awesome. LGTM. 🎉

src/service-override/files.ts
CGNonofr: Found it while investigating the issue you've reported!

src/service-override/files.ts
```diff
@@ -194,8 +194,9 @@ class RegisteredFileSystemProvider extends Disposable implements IFileSystemProv
 
   public async readdir (resource: URI): Promise<[string, FileType][]> {
     const includedPaths = Array.from(this.files.keys())
-      .map(uri => extUri.relativePath(resource, URI.parse(uri))!)
-      .filter(path => !path.startsWith('..'))
+      .map(uri => URI.parse(uri))
+      .filter(uri => uri.path.startsWith(resource.path))
```
CompuIves, line 198: I think this can be a problematic check. For example, if the directory is `/a`, then all files that start with `/a` will be included. I think the check should be something along the lines of:

```
.filter(uri => uri.path.startsWith(resource.path + '/'))
```

I've been bitten by a check like this before, maybe it does not apply here.

src/service-override/files.ts
kaisalmen: Just curious. How did you identify this? Debugging in Chrome or what did you do? I had no starting point for analysing it.

CGNonofr: It was surprisingly easy! I just had to "pause script execution" in chrome dev tools to realize it was calling this method in loop recursively until a stack overflow occurs and then do it again

PR #429, opened 2024-05-10
fix: add missing accessibleView service

CompuIves:
I noticed that sometimes a new service (introduced with VSCode 1.89) is required, called the `AccessibleViewInformationService`. In this PR I add this service, together with another service related to accessibility I found.

Fixes: 

image

### DIFF
--- a/src/missing-services.ts
+++ b/src/missing-services.ts
@@ -222,6 +222,7 @@ import { AccountStatus } from 'vs/workbench/services/userDataSync/common/userDat
 import { IWebview } from 'vs/workbench/contrib/webview/browser/webview'
 import { SyncResource } from 'vs/workbench/contrib/editSessions/common/editSessions'
 import { ILanguageModelStatsService } from 'vs/workbench/contrib/chat/common/languageModelStats.service'
+import { IAccessibleViewInformationService } from 'vs/workbench/services/accessibility/common/accessibleViewInformationService.service'
 import { getBuiltInExtensionTranslationsUris, getExtensionIdProvidingCurrentLocale } from './l10n'
 import { unsupported } from './tools'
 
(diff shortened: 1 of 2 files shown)


src/missing-services.ts
```diff
@@ -2290,6 +2292,24 @@ registerSingleton(IAccessibleViewService, class AccessibleViewService implements
...
+  onDidChangeReducedMotion = Event.None
+  alwaysUnderlineAccessKeys = async () => false
+  isScreenReaderOptimized = () => false
+  isMotionReduced = () => false
+  getAccessibilitySupport = () => 0
+  setAccessibilitySupport = unsupported
+  alert = unsupported
+  status = unsupported
+}, InstantiationType.Delayed)
+
```
CGNonofr, line 2312: This service identifier is already registered with the real impl in the editor standalone services. It probably means the editor needs it, so registering it twice with a mocked impl is dangerous: if a race condition make this code execute before the standaloneServices registration, this empty impl will be used and it may break things

CGNonofr (changes requested):
Thank you!

src/missing-services.ts
CompuIves: Oh good one, removed it!

src/service-override/accessibility.ts
```diff
@@ -15,6 +19,8 @@ registerAssets(audioAssets)
 export default function getServiceOverride (): IEditorOverrideServices {
   return {
     [IAccessibleViewService.toString()]: new SyncDescriptor(AccessibleViewService, [], true),
-    [IAccessibilitySignalService.toString()]: new SyncDescriptor(AccessibilitySignalService, [], true)
+    [IAccessibilitySignalService.toString()]: new SyncDescriptor(AccessibilitySignalService, [], true),
+    [IAccessibilityService.toString()]: new SyncDescriptor(AccessibilityService, [], true),
```
CGNonofr, line 23: It seems superfluous but meh, it can't hurt I guess!

CGNonofr (approved):

src/service-override/accessibility.ts
CompuIves: Who knows if they'll add some logic in the future! 😂

CompuIves:
Once the checks succeed, would it be possible to cut a new release? I think this is the main thing holding us back from upgrading to the new VSCode (for some reason, I think that sidebar views contributed by extensions don't work because this service is missing)

CGNonofr:
> Once the checks succeed, would it be possible to cut a new release? I think this is the main thing holding us back from upgrading to the new VSCode (for some reason, I think that sidebar views contributed by extensions don't work because this service is missing)

Sure, I would just like to add some small fixes (PR incoming soon)

PR #434, opened 2024-05-23
Improve registered file system

CGNonofr:
What it improves:
- it use a tree to store the filesystem instead of a flat file list (for better perfs, similiar to InMemoryFileSystemProvider)
- it implement the file stream api
- it doesn't rely anymore on the file content to get the file size => removing the need to get the file twice (one from the `stat` to get the size then another one to get the content) => removing the need to `memoize` the file content => stop keeping all the file content in memory, reducing memory footprint


kaisalmen (approved):
Sounds good. As usual I will test the new release with the `monaco-languageclient` stack. LGTM.

PR #445, opened 2024-06-06
Update to VSCode 1.90

CGNonofr:


kaisalmen:
@CGNonofr I tested it locally and everything seems fine, but you broke the demo build

kaisalmen:
I will give you a go once that is fixed.

kaisalmen:
@CGNonofr I just read this in the VSCode release notes:
image

Does it mean you can/should move to node 20 here as well?

CGNonofr:
> @CGNonofr I just read this in the VSCode release notes: image
> 
> Does it mean you can/should move to node 20 here as well?

AFAIK, They seem to be still using node 18 for building though, but I've switched to last versions and everything seems to work so... updated

CGNonofr:
> > @CGNonofr I just read this in the VSCode release notes: image
> > Does it mean you can/should move to node 20 here as well?
> 
> AFAIK, They seem to be still using node 18 for building though, but I've switched to last versions and everything seems to work so... updated

Hum nop, doesn't work, rollbacked

kaisalmen (approved):
LGTM

kaisalmen:
> Hum nop, doesn't work, rollbacked

ok, not aligning build and execution env sounds fishy. On the other hand the SW stack involved is deep and broad regarding dependencies.

PR #475, opened 2024-08-02
Update to VSCode 1.92

CGNonofr:
Also microsoft did change a lot how language packages work, allowing us to removing a lot of code and be way closer to how it works in VSCode itself by using the same bundled format (which is lighter because it contains only indexes instead of string keys)


kaisalmen (approved):
Thank you. LGTM. I will test and integrate the new release with monaco-languageclient and its examples.

CGNonofr:
@kaisalmen forgot to make it as a breaking change :(

kaisalmen:
@CGNonofr can you re-release it as 8.0.0?

CGNonofr:
> @CGNonofr can you re-release it as 8.0.0?

done!

PR #499, opened 2024-09-17
Use workbench editor worker service

CGNonofr:
The worker name was updated only in the workbench service we don't use yet


kaisalmen (approved):
LGTM

PR #505, opened 2024-10-08
Fix demo in google chrome

CGNonofr:
For some reasons, the demo seems to be not working anymore, only in google chrome

I'm unable to get any error, it looks like a syntax error in a worker?

Anyway, disabling `minifySyntax` solves the issue :shrug: 

already deployed to [netlify](https://monaco-vscode-api.netlify.app/)


kaisalmen (approved):
LGTM. Thank you

PR #516, opened 2024-10-19
Do not throw error from IEnvironmentVariableService

CGNonofr:


kaisalmen (approved):
LGTM. Just curious. What was the reason for using the annotations?

CGNonofr:
> LGTM. Just curious. What was the reason for using the annotations?

I want the error message to contain the service name, do you have an alternative?

kaisalmen:
No, I was first confused why there were some many lines changed.

CGNonofr:
> No, I was first confused why there were some many lines changed.

In typescript, you can't have annotations on classes with expression is immediately used. I should probably have refactored it in another commit :thinking: sorry

PR #527, opened 2024-10-29
Fix worker configuration, for real

CGNonofr:
I tried to homogenize the imports but it was a mistake...

`editorSimpleWorker.esm.js` doesn't reexport the `initialize` function


kaisalmen (approved):
LGTM

PR #520, opened 2024-10-22
Fix main dependencies

CGNonofr:
fix https://github.com/CodinGame/monaco-vscode-api/discussions/518

PR #532, opened 2024-11-06
Update to VSCode 1.95

CGNonofr:
When upgrading, the https://github.com/CodinGame/monaco-vscode-api/pull/520 issue came back

So to fix it, I've changed the build behavior: instead of putting everything that is used by at least 2 service-override in the main package, I create a `common` package for it. To not create too many new packages, only common packages containing at least 10 modules are kept, others still rely on the previous behavior


CGNonofr:
@kaisalmen build fixed :)

kaisalmen (approved):
Thank you. That is a bigger change, but it makes sense. I will check with monaco-languageclient examples as usual. LGTM

CGNonofr:
> Thank you. That is a bigger change, but it makes sense. I will check with monaco-languageclient examples as usual. LGTM

It's a big change indeed that was pretty challenging to implement, the rollup code is now quite complex :thinking: I'll try to rework it to simplify it some day, it probably still contain a lot of legacy stuff

PR #543, opened 2024-12-13
Update rollup.config.ts to expose contentHoverController

rubenfiszel:


rubenfiszel:
Used by monaco-graphql (after vite aliasing). 
Was exposed in a previous commit but under just .../hoverController and got deleted recently which break compatibility with monaco-graphql

CGNonofr (approved):

PR #546, opened 2024-12-20
fix: bunle everything if the extension has a "main" entry point

mame:
VScode extensions with `main` entry point contain some js, as well as `browser`.

I tried bundling [vscode-rbs-syntax vsix](https://github.com/soutaro/vscode-rbs-syntax) but `rollup-vsix-plugin` did not include its `out/extension.js`. If I specify a dummy `browser` entry point to the vsix, it worked. So I think this change is needed.

### DIFF
--- a/src/extension-tools.ts
+++ b/src/extension-tools.ts
@@ -220,7 +220,7 @@ export async function getExtensionResources(
   fs: typeof nodeFs,
   cwd: string
 ): Promise<ExtensionResource[]> {
-  if (manifest.browser != null || manifest.contributes?.typescriptServerPlugins != null) {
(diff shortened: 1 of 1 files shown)


CGNonofr:
The extension declaring a `main` field without a `browser` field means it's not browser compatible (try to install it in vscode.dev). It it's able to run in the browser, it's the extension responsibility to declare a browser field.

I'll close the PR.

You should probably send a PR on their side instead.

In the meantime, if you're confident the extension is browser-compatible, the vsix plugin takes a parameter allowing to rewrite the manifest so you can add the browser entry on the fly.

mame:
I see, I will talk with the maintainer of vscode-rbs-syntax. Thank you for your time!

PR #555, opened 2025-01-20
Revamp build process and update to 1.96.3

CGNonofr:
I'm aware it's a HUGE PR, sorry for that, I don't really expect you to fully understand what has been done about the build process as it would be really time consuming.

It's tried to fix all the issues I've noticed about the build process and decided to fully rework it on healthier bases.

What was improved:
- The packages dependencies should now be properly set on all the subpackages
- Each vscode module now has its .d.ts next to it, which allows to import any vscode file from the final project with the types
- It's faster (there is now only 2 rounds of rollup instead of 3, and no separate types bundling)

Regarding the type, I had to make a choice, because the import graph of types and the one from the actual ESM are different (for instance, if only a type is imported from another module, the import is removed from the generated ESM).

I've decided that only external dependencies imported from actual ESM files are declared in the package `dependencies`. External dependencies only imported from d.ts files are put in the `peerDependencies` as optional (using `peerDependenciesMeta`). Any though about it?

Also, I've upgraded VSCode in the same MR so a new major release is triggered in case something was broken in the process.

And whilte I'm at it, I've updated every single dependencies, including eslint which I've switch to the default recommended configuration for now


CGNonofr:
@kaisalmen do you have an idea about the build error? I've tried many things without any differences :thinking: 

No issue on my machine of course

kaisalmen:
@CGNonofr awesome. I will look at all the changes later today.

> do you have an idea about the build error?

I am locally on Ubunut 24.04.1 and it fails at the same point with a slightly different reason. I will come back to you...

kaisalmen:
@CGNonofr now it fails here:
image

I will install ajv as devDependency and see what happens

kaisalmen:
@CGNonofr if yarn is still needed you should also enforce the version with volta

CGNonofr:
> I am locally on Ubunut 24.04.1 and it fails at the same point with a slightly different reason. I will come back to you...

Interesting!

> I upgraded volta npm to 22.13.0 (current LTS) and it works locally now so far. It went further but is not yet completed.

If you look at the commit history, you would notice that I've tried playing with the node/npm version already (:

> if yarn is still needed you should also enforce the version with volta

I was thinking the whole repo has switched from yarn to npm, but maybe not everywhere? :thinking:

kaisalmen:
@CGNonofr adding ajv did not help

kaisalmen:
> I was thinking the whole repo has switched from yarn to npm, but maybe not everywhere? 🤔

There are two place left in `install-vscode`

kaisalmen:
@CGNonofr you can replace yarn with npx and the package.json in the tmp directory needs to contain `ajv` and `ajv-keywords`. It works when I change that manually, so I assume the package.json needs to be patched, too.

CGNonofr:
> @CGNonofr you can replace yarn with npx in `install-vscode` and the package.json in the tmp directory needs to contain `ajv` and `ajv-keywords`. It works when I change that manually, so I assume the package.json needs to be patched, too.

Can you push the changes you mention? I'm not sure to fully understand

kaisalmen:
Done. ⬆️ How supplies the package.json in the /tmp directory? Is that part of the vscode download? The package.json in the tmp directory needs to contain `ajv` and `ajv-keywords`

CGNonofr:
> Done. ⬆️ How supplies the package.json in the /tmp directory? Is that part of the vscode download? The package.json in the tmp directory needs to contain `ajv` and `ajv-keywords`

That package.json is just the package.json of the VSCode repo. The only way to update it is via the patches

kaisalmen:
@CGNonofr node-pty works on my machine. There must  be a stupid difference. I have the exact same python version locally

kaisalmen:
@CGNonofr One more try: I install additional libraries

kaisalmen:
@CGNonofr that is what is missing, I guess / or at least some:
https://github.com/microsoft/vscode/blob/main/.github/workflows/ci.yml#L111

kaisalmen:
Yes, it now got past that error

CGNonofr:
> Yes, it now got past that error

nice! there is another error about ajv though :thinking:

kaisalmen:
> nice! there is another error about ajv though 🤔

That is the same I see locally. Can you path the vscode package.json to contain those package (`ajv` and `ajv-keywords`)?

CGNonofr:
@kaisalmen this time I'm confident :)

kaisalmen:
> this time I'm confident :) 

🤞

CGNonofr:
> > this time I'm confident :)
> 
> 🤞

demo failed !!

CGNonofr:
@kaisalmen All green :tada:

kaisalmen:
> All green 🎉 Thanks for your help!

Wonderful. Did you already test the release process?

kaisalmen (approved):
Let's go. We have to test this in the field. 🎉

CGNonofr:
> > All green 🎉 Thanks for your help!
> 
> Wonderful. Did you already test the release process?

It's always hard to test that, but the only difference is the subpackage path so I hope nothing broke 🤞

kaisalmen:
@CGNonofr I saw that the creation of releases failed. Can you ping me once they are available? Thank you.

CGNonofr:
@kaisalmen You were already pinged :)

kaisalmen:
@CGNonofr dependency problems, see: https://github.com/CodinGame/monaco-vscode-api/issues/557

Issue #563, opened 2025-01-27
Dependency issues with v13 / questions


kaisalmen:
`@codingame/monaco-vscode-extension-api` has `@types/vscode` as peerDependency. This should be devDependency.

Are you aware there is a dependency loop from `@codingame/monaco-vscode-api` => `@codingame/monaco-vscode-files-service-override` => `@codingame/monaco-vscode-api` and other packages?

Is it still necessary to redefine `monaco-editor` on this level or could you use `@codingame/monaco-vscode-editor-api` everywhere? I no longer do this  on `monaco-languageclient` level (https://github.com/TypeFox/monaco-languageclient/pull/836).

Would this be possible with vscode as well or is the global package rewrite required?

Btw, we currently cannot prevent
```yaml
  "dependencies": {
    "@codingame/monaco-vscode-api": "13.0.0",
    "@codingame/monaco-vscode-model-service-override": "13.0.0",
    "@codingame/monaco-vscode-extensions-service-override": "^10"
  }
```
to install both versions? I always thought that the closest version of `@codingame/monaco-vscode-api` drives the deduplication. But, it seems the problem comes from `vscode@npm:@codingame/monaco-vscode-api@10.1.4` being the previous dependency to `@codingame/monaco-vscode-extensions-service-override`. So, we can only fix this from now on (>=13) and the old workaround for monaco-languageclient should be kept in place for past versions.

Does it make sense to define only define `@codingame/monaco-vscode-api` as peer dependency and everything else either as regular or [optional dependency](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#optionaldependencies) to remove complexity?

kaisalmen:
@CGNonofr you can use these projects for playing around. They are independent of the workspace:
http://github.com/TypeFox/monaco-languageclient/tree/mva-13/verify/peerNpm
http://github.com/TypeFox/monaco-languageclient/tree/mva-13/verify/peerPnpm
http://github.com/TypeFox/monaco-languageclient/tree/mva-13/verify/peerYarn

CGNonofr:
> @codingame/monaco-vscode-extension-api has @types/vscode as peerDependency. This should be devDependency.

devDependencies have no consequences for published packages, so might as well delete it. Why is it an issue?

> Are you aware there is a dependency loop from @codingame/monaco-vscode-api => @codingame/monaco-vscode-files-service-override => @codingame/monaco-vscode-api and other packages?

Yes, is it an issue?

[...]

PR #564, opened 2025-01-27
Fix dependency issues (and other fixes)

CGNonofr:
fix https://github.com/CodinGame/monaco-vscode-api/issues/563


CGNonofr:
> The monacoVscodeApiInitialized is simple and effective

Unfortunately it will only work for new versions though!

PR #570, opened 2025-02-03
Fix type export

CGNonofr:
Just a few minor fixes


kaisalmen (approved):
Sorry, I have overlooked that yesterday. LGTM

PR #571, opened 2025-02-04
Wrong type dedup

CGNonofr:
before:
image
image

after:
image
image

(missing export, and wrong monaco-editor path)

(it's red just because the screenshot is taken from the lib dist folder)


kaisalmen (approved):
LGTM

PR #580, opened 2025-02-11
Allow configuring extension host worker options

CGNonofr:
It was removed during a VSCode refactor removing non-ESM code. But we do need to support non-esm workers for webpack

### DIFF
--- a/src/monaco.ts
+++ b/src/monaco.ts
@@ -81,6 +81,7 @@ export {
   EditorCommand
 } from 'vs/editor/browser/editorExtensions'
 export type { IEditorContribution, IDiffEditorContribution } from 'vs/editor/common/editorCommon'
+export type { ITextFileEditorModelSaveEvent } from 'vs/workbench/services/textfile/common/textfiles'
 
 function computeConfiguration(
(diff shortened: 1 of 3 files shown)


kaisalmen (approved):
Then let's bring the options back. LGTM

CGNonofr:
> Then let's bring the options back. LGTM

Don't you have any issue regarding it on your side?

kaisalmen:
@CGNonofr we need to adapt.

kaisalmen:
The v14 PR in mlc is still a Draft 🙂

CGNonofr:
> The v14 PR in mlc is still a Draft 🙂

It's not a regression of the v14 though, it's been multiple versions

CGNonofr:
> Yes, I understand. It is fixed to a module as well: https://github.com/TypeFox/monaco-languageclient/blob/main/packages/client/src/vscode/services.ts#L163-L175

webpack removes  the  `type: 'module` though

kaisalmen:
Fancy piece of software! 😉  Yes, that must be made configurable there ⬆️

PR #581, opened 2025-02-11
Fix extension host worker when used with a cdn

CGNonofr:
If the extension host iframe is hosted on another domain, it refuses messages from the main thread because the domain does not match. VSCode handle this by adding a `parentOrigin` parameter but in our case it wasn't used

Also, force set the domain of the extension host worker, it was the case before a previous refactor and is a regression, forcing us to set the webpack `publicPath` for it to work, or provide a `iframeAlternateDomain` by hands

### DIFF
--- a/src/service-override/extensions.ts
+++ b/src/service-override/extensions.ts
@@ -291,7 +291,7 @@ class ExtensionResourceLoaderServiceOverride extends ExtensionResourceLoaderServ
   }
 }
 
-let iframeAlternateDomains: string | undefined
+let iframeAlternateDomain: string | undefined
 registerAssets({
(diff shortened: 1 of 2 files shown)


CGNonofr:
@kaisalmen I'm trying to update the lib in our apps which all use the last version of webpack, and I'm surprised to encounter so many issues that nobody had before :thinking:

kaisalmen (approved):
LGTM

kaisalmen:
> and I'm surprised to encounter so many issues that nobody had before 🤔

Webpack changed to better or worse or nobody looked closely enough. I don't use webpack if I don't have to, so the checks in mlc are limited to basic bundling checks.

PR #586, opened 2025-02-12
remove unwanted classes from body

Daniil8k:
There is no need to add extra classes to the body, as these classes are specific to the editor area.
Otherwise, they pollute the `body` unless you attach the editor directly below the body.

So we only need to attach 'web' and platformClass to `mainContainer`. And that is already done in the line above.

### DIFF
--- a/src/service-override/layout.ts
+++ b/src/service-override/layout.ts
@@ -70,8 +70,6 @@ export class LayoutService extends Disposable implements ILayoutService, IWorkbe
     ])
 
     mainContainer.classList.add(...workbenchClasses)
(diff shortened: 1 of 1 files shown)


CGNonofr:
> As I see in codingame.com, 'web' or platformClass on body doesn't make any sense btw. Because those fonts are fully dropped in [this patch](https://github.com/CodinGame/monaco-vscode-api/blob/main/vscode-patches/0025-feat-only-apply-style-on-specific-class.patch).

That's a good point actually :thinking: I'm convinced!

Daniil8k:
Thanks 😁

PR #591, opened 2025-02-27
Fix export order

### LINKED ISSUE #590: Incorrect exports Order in package.json Causes Webpack Build Error
The latest versions of certain packages, such as:
- monaco-vscode-api
- monaco-vscode-quickaccess-service-override
- monaco-vscode-layout-service-override
- etc.

contain an incorrect exports field in their package.json file. Specifically, the "default" condition appears at the beginning of the list instead of at the end.

Issue
When using Webpack, this results in the following build-time error:

`Module not found: Error: Default condition should be last one`

Proposed Solution
The order of the exports field should be fixed so that "default" is listed last, as per Node.js module resolution rules. Example correction:

✅ Correct Order:

```
"exports": {
    ".": {
      "services": "./services.d.ts",
      "default": "./services.js"
    },
    "./vscode/*": {
      "types": "./vscode/src/*.d.ts",
      "default": "./vscode/src/*.js"
    },
    "./*": {
      "types": "./*.d.ts",
      "default": "./*.js"
    }
  },
```

🚫 Incorrect Order (Current Issue):
```
"exports": {
    ".": {
      "default": "./services.js",
      "services": "./services.d.ts"
    },
    "./vscode/*": {
      "default": "./vscode/src/*.js",
      "types": "./vscode/src/*.d.ts"
    },
    "./*": {
      "default": "./*.js",
      "types": "./*.d.ts"
    }
  },
```

### Steps to Reproduce

1. Install the latest version of the affected package(s).
2. Attempt to build a Webpack project that imports these packages.
3. Observe the error message: "Default condition should be last one"

### Expected Behavior
The packages should define exports correctly to prevent Webpack build failures.

### Environment Details
Webpack Version: 5.78.0
Affected Package Versions: 14.0.2 and higher, worked fine with 11.1.2
Node.js Version: 22.14.0

Would appreciate a fix in an upcoming release! Thanks!

CGNonofr:
fix  #590

### DIFF
--- a/rollup/tools/configuredSubpackagePlugin.ts
+++ b/rollup/tools/configuredSubpackagePlugin.ts
@@ -371,16 +371,16 @@ ${code}`
             module: 'services.js',
             exports: {
               '.': {
-                default: './services.js',
-                services: './services.d.ts'
+                services: './services.d.ts',
+                default: './services.js'
               },
               './vscode/*': {
-                default: './vscode/src/*.js',
-                types: './vscode/src/*.d.ts'
+                types: './vscode/src/*.d.ts',
+                default: './vscode/src/*.js'
               },
               './*': {
-                default: './*.js',
-                types: './*.d.ts'
+                types: './*.d.ts',
+                default: './*.js'
               }
             },
             typesVersions: {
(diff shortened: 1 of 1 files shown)


rollup/tools/configuredSubpackagePlugin.ts
```diff
@@ -371,16 +371,16 @@ ${code}`
             module: 'services.js',
             exports: {
               '.': {
-                default: './services.js',
-                services: './services.d.ts'
+                services: './services.d.ts',
```
kaisalmen, line 374: That must be `types`

kaisalmen (commented):
Is the typo the actual problem and not the order?

CGNonofr:
> Is the typo the actual problem and not the order?

The error message is pretty clean though

In my experience, the type `typesVersions` field is used and the exports.types is pretty much ignored :thinking: , which explain why we didn't notice the typo

kaisalmen (approved):
LGTM

PR #592, opened 2025-03-03
Fix quickinput standalone position

CGNonofr:
Microsoft doesn't seem to be releasing monaco-editor anymore :thinking: 

They broke the standalone quickinput location while implementing the ability to move it in the workbench

I'm not sure what is the proper fix to it, but at least  that MR seems to be restoring the previous behavior


kaisalmen (dismissed):
Interesting, there aren't even any next builds. LGTM

CGNonofr:
@kaisalmen forgot to remove the import as well, wasn't building...

kaisalmen (approved):
LGTM once build is green.

PR #597, opened 2025-03-10
Fix completely wrong update

CGNonofr:
I was sure I was able to test it but it looks like I did something wrong. The demo is currently not working at all for multiple reasons.


kaisalmen (approved):
LGTM. I will test it with the monaco-languageclient stack tomorrow.

PR #605, opened 2025-04-04
VSCode 1.99.0

CGNonofr:
Also change the build process a bit, so that CSS file imports and their assets are kept in the bundle instead of being inlined, allowing the final project to decide what to do with the style, use postcss one it...


kaisalmen (approved):
Looks good. Let's see if the css non-inline change makes any problems down the line... I will provide feedback as soon as I have checked it. LGTM

PR #609, opened 2025-04-10
export initialize of workers

erickRochaIP:


CGNonofr:
Thanks for your contribution!

It appears that this file is used for both the editor worker, and the standalone worker, so your change is probably breaking the main editor worker

CGNonofr:
The api was changed more drastically, and the fix is not that simple. It was implemented in https://github.com/CodinGame/monaco-vscode-api/pull/610

PR #616, opened 2025-04-24
Move webview html next to service worker

CGNonofr:
fix #615

Also synchronize the html file from the original


kaisalmen (approved):
LGTM once the build is green.

PR #620, opened 2025-05-14
Vscode 1.100.1

CGNonofr:
Also support overriding the base product configuration using the `_VSCODE_PRODUCT_JSON` global variable

Some code from VSCode relies on it and ignore the product override, like the default chat configuration


kaisalmen (approved):
Thank you. LGTM

PR #627, opened 2025-05-27
Depend on dompurify instead of using inlined version

CGNonofr:
Just do the same as what as been done for marked to dompurify

For some reasons, the VSCode repo includes the built version of some libraries instead of just referencing them

It has cons for us:
- If the final project already uses the library, the code is duplicated
- The included libraries are in cjs format, while they are available in ESM in their published version
- It prevents the detection of vulnerable versions of dependencies

### DIFF
--- a/demo/package-lock.json
+++ b/demo/package-lock.json
@@ -732,6 +732,7 @@
       "license": "MIT",
       "dependencies": {
         "@codingame/monaco-vscode-api": "0.0.0-semantic-release",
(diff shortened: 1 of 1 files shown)


CGNonofr:
> Makes sense. LGTM once the build is green. Do you have any idea if there are any more contained/hidden from package management libraries?

I don't think so, the only 3 seems to be
- dompurify
- marked
- semver

PR #637, opened 2025-06-16
fix(overlayfs): mkdirp on all write requests

### LINKED ISSUE #636: Explorer View: OverlayFileSystemProvider mkdir fails
This can be reproduced with the official demo or the [mlc application example](https://typefox.github.io/monaco-languageclient/ghp_appPlayground.html). When you try to create a new directory with the corresponding button from the Explorer View you get a NoPermission error ⬇ . Interestingly, this does not happen if you create a file. More interestingly, after the successful file creation the directory creation suddenly works as well with any error.
Image

CGNonofr:
fix #636

### DIFF
--- a/src/service-override/files.ts
+++ b/src/service-override/files.ts
@@ -792,6 +792,7 @@ class OverlayFileSystemProvider
   }
 
   private async writeToDelegates(
+    resource: URI,
     caller: (delegate: IFileSystemProviderWithFileReadWriteCapability) => Promise<void>
   ): Promise<void> {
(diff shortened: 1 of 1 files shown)


kaisalmen (approved):
Thank you. LGTM

CGNonofr:
> > Btw, do you need that fix to be released?
> 
> If you are anyhow releasing v18.0.0 this week there is no need. If not, then create a bugfix release, please. Thank you.

Probably not, I'll trigger a release

PR #648, opened 2025-06-26
Properly pull BrowserWorkingCopyBackupTracker

CGNonofr:
[regression] the registration of the BrowserWorkingCopyBackupTracker, which is done inside browser/workingCopyBackupService.ts, is not done anymore, so working copy of unsaved files aren't saved anymore


onora-hbl (approved):
Looks good, thanks for the explanation

PR #653, opened 2025-07-07
Register language pack extension

CGNonofr:
Not a big deal, allows to remove this button from the interface (when using a server that has the language pack, and the client too):
image

### DIFF
--- a/src/services.ts
+++ b/src/services.ts
@@ -350,3 +350,283 @@ export type {
   ServicesAccessor
 }
 export { ConfigurationTarget } from 'vs/platform/configuration/common/configuration'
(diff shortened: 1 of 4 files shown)


src/services.ts
```diff
@@ -350,3 +350,283 @@ export type {
   ServicesAccessor
 }
 export { ConfigurationTarget } from 'vs/platform/configuration/common/configuration'
+export { IRemoteExplorerService } from 'vs/workbench/services/remote/common/remoteExplorerService.service'
```
kaisalmen, line 353: Home come there are so many new exports?

CGNonofr: Did you check the commit message?

PR #682, opened 2025-08-21
feat: Add VSCode File System Search Integration

aliadelelroby:
- Implement modular search architecture with separate utilities
- Add complete directory traversal and indexing capabilities
- Integrate with VSCode file system APIs for real-time updates
- Include scalability optimizations for large projects
- Add comprehensive error handling and validation
- Support for both file and text search providers
- Automatic exclusion of build directories and common artifacts
- Memory and timeout limits to prevent resource exhaustion

Files:
- demo/src/features/search-utils/ (new modular utilities)
- demo/src/features/search.ts (refactored main search file)


demo/src/features/search-utils/errors.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  static warn(message: string, ...args: unknown[]): void {
+    console.warn(`[Search] ${message}`, ...args)
+  }
+
+  static error(message: string, error?: Error, ...args: unknown[]): void {
+    console.error(`[Search] ${message}`, error, ...args)
+  }
+}
```
CGNonofr, line 40: What about using a vscode logger instead?

demo/src/features/search-utils/base-provider.ts
```diff
@@ -0,0 +1,301 @@
...
+
+  protected isWithinSizeLimit(uri: vscode.Uri): Promise<boolean> {
+    return new Promise((resolve) => {
+      this.fileService
+        .resolve(uri)
+        .then((stats) => {
+          resolve((stats.size ?? 0) <= this.config.maxFileSize)
+        })
+        .catch(() => resolve(false))
+    })
```
CGNonofr, line 299: I'm confused why you didn't just declare the method async and await/try-catch `this.fileService.resolve` :thinking:

CGNonofr (commented):
That's a LOT of code! Does it come from somewhere else? Otherwise you're very fast to write code (:

CGNonofr:
> All the pattern-matching code comes directly from VS Code; I only added the necessary parts to ensure it matches VS Code’s functionality.

I see :thinking: That's a shame to copy/paste all that code from VSCode though :thinking: 

What about returning those providers from the search service override directly? so that you can directly import from VSCode source code, instead of duplicating it?

aliadelelroby:
You're right! I'll refactor it right now to use the VSCode service directly instead.

CGNonofr:
> Done, now we're using the pattern matching code directly from the VSCode search override.

Better ! :)

What about exporting those provider from the search service override though? so it can be used by anyone? I wasn't expecting this to be so complex :)

CGNonofr:
> Do you mean adding them to the searchServiceOverride in **monaco-vscode-api** (not just the demo) and exporting them from there?

exactly :)

And while you're at it, it would be also very nice to make it work using ISearchService.registerSearchResultProvider directly, instead of using the extension api :)

aliadelelroby:
Hey! I just added it to the search service override by default if **isHTMLFileSystemProvider** result is false.

src/service-override/tools/search-utils/types.ts
```diff
@@ -0,0 +1,32 @@
...
+    '**/*.tmp',
+    '**/*.temp',
+    '**/*.log'
+  ],
+  maxFileSize: 1024 * 1024 * 10,
+  debounceDelay: 300
+}
+
+export const MAX_CACHED_FILES = 100000
+export const MAX_DIRECTORY_DEPTH = 10
```
CGNonofr, line 32: I'm consufed by the file name `types` since it contains constants :thinking:

src/service-override/tools/search-utils/types.ts
```diff
@@ -0,0 +1,32 @@
...
+    '**/dist/**',
+    '**/build/**',
+    '**/out/**',
+    '**/*.tsbuildinfo',
+    '**/*.code-workspace',
+    '**/.DS_Store',
+    '**/*.tmp',
+    '**/*.temp',
+    '**/*.log'
+  ],
```
CGNonofr, line 26: shouldn't it be using the config `files.exclude` instead?

aliadelelroby:
Oh yeah, you're right about the remote file system, I just made a commit addressing it.

rollup/rollup.config.ts
```diff
@@ -94,6 +94,10 @@ export default (args: Record<string, string>): rollup.RollupOptions => {
         preserveModules: true,
         preserveModulesRoot: SRC_DIR,
         minifyInternalExports: false,
+        sanitizeFileName(fileName) {
+          // Remove spaces in name to prevent creating any issues
+          return fileName.replace(/\s+/g, '_')
+        },
```
CGNonofr, line 100: Not sure what I'm looking at! did you squash a merge commit of something?

PR #686, opened 2025-08-25
feat: add workspace search with VSCode FS

aliadelelroby:
This PR implements a full workspace search provider integrated with the VSCode file system, enabling efficient file searching across the entire workspace. It includes updates to support dynamic file indexing and it's fully integrated with the existing file system APIs.


aliadelelroby:
@CGNonofr  Is it good now ?

CGNonofr:
Looks good for a first version :) my main concern is that everything runs in the main thread so it may freeze the browser during the search :thinking: but probably not a big deal since there will probably not be a lot of files there

CGNonofr (approved):
Thanks! well done. I didn't review the implementation in details but we will be able to iterate from there :)

PR #688, opened 2025-09-05
Fix extension bundle content loading when using data url

CGNonofr:
[VSCode fetch the extension l10n bundles via the FileService](https://github.com/microsoft/vscode/blob/67eb75414b30401ac261031f7e47f7685a9a97bd/src/vs/workbench/api/browser/mainThreadLocalization.ts#L33-L36) and those file can be bundled as base64 data uris

the FileService currently fails to load them, which leads to default extension not being translated when the bundler decides to use data urls for those files

### DIFF
--- a/src/service-override/files.ts
+++ b/src/service-override/files.ts
@@ -1094,6 +1094,16 @@ class FileServiceOverride extends FileService {
     }
   }
 
+  override async withProvider(resource: URI) {
+    if (resource.scheme === 'data') {
+      const httpProvider = this.getProvider('http')
+      if (httpProvider != null) {
+        return httpProvider
+      }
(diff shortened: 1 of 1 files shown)


onora-hbl (approved):
Looks good

CGNonofr:
> Not sure if this is related, but I am getting following error when bundled using webpack. I am running in a single page application. Here is how I am loading the .json files in extension: https://github.com/dhrubomoy/monaco-languageclient/blob/0271e9f652a2ef77f0ff868bd393fee33037718e/packages/examples/src/langium/statemachine/config/statemachineConfig.ts#L86-L89

Can you please open your own issue?

PR #709, opened 2025-09-29
Prevent VSCode from breaking default extension sourcemap references

CGNonofr:
The VSCode default extension build process replaces the local sourcemap references by links to `https://main.vscode-cdn.net/sourcemaps/XXX`

It's an issue because:
- It's currently broken because the url contains `undefined` instead of the VSCode commit
- We may have patched the default extensions and the published sourcemap may not be corresponding to the code

I'm not sure why microsoft is doing such transformation :shrug: 

The change prevent vite from logging a lot of sourcemap errors


kaisalmen (approved):
LGTM. I will test it with mlc.

PR #753, opened 2025-12-15
VSCode 1.107

CGNonofr:


CGNonofr:
> @CGNonofr so now we can override where the extension host worker is loaded from?

Can you explain a bit what you mean? do you mean in the code? or the url? I don't expect it to be more than a refactor :thinking:

kaisalmen:
There is a default, but you can override it, of course. That is done in one or two example and the mentioned tests. I think I need to adjust the above code.

PR #754, opened 2025-12-18
Refactor common packages

CGNonofr:
stop generating *-common packages with uuids

Instead, try to merge them as much as possible with each others and with the main package, at the only condition that it doesn't add additional external dependencies (xterm, xterm addons, katex...)

Name the common packages with the external dependencies they are pulling (instead of uuids) so it's more predictable

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,9 +30,9 @@
   },
   "config": {
     "vscode": {
-      "version": "1.107.0",
-      "ref": "1.107.0",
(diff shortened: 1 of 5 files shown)


kaisalmen:
@CGNonofr wow, I didn't expect such a change. That sounds good. How confident are you it is working properly?
Do you want to trigger a pre-release this time?

Issue #804, opened 2026-06-03
Webview messages from a remote extension stall after the first round-trip — openai.chatgpt (Codex) SPA never finishes bootstrap


nikolai-vysotskyi:
### Summary

When hosting a webview-heavy extension as a **remote** extension (`ExtensionHostKind.Remote` + `getRemoteAgentServiceOverride({ scanRemoteExtensions: true })` + a real VS Code server), the extension's webview SPA loads fully but **stalls mid-bootstrap**: the first webview↔extension round-trip succeeds, then no further messages flow, so the SPA never reaches its "ready" state. The **same `.vsix` works perfectly in stock VS Code**, so the extension itself is fine.

Concrete case: OpenAI's `openai.chatgpt` (Codex) extension.

### Environment

- `@codingame/monaco-vscode-api` **33.0.9**
- VS Code server pinned to commit `987c9597516278c9fcf10d963a0592ce1384ab93`
- Remote extension host (REH server) on `localhost:<port>`, `remoteAuthority` set
- Electron 41, macOS (arm64), React 19 renderer
- Service overrides: configuration, extensions, files, model, theme, view-common, views, storage, secret-storage, authentication, notifications, dialogs, chat, extension-gallery, remote-agent

### Repro

1. Install `openai.chatgpt` (platform `darwin-arm64`) into the REH server's extensions dir; enable its proposed APIs (`chatSessionsProvider`, `languageModelProxy`) via `product.json`.
2. Boot monaco-vscode-api as a remote client to the server; open Codex's custom editor (`chatgpt.conversationEditor`, scheme `openai-codex:`) or its webview view.

### Expected

SPA boots and renders the chat UI (it does in stock VS Code — verified: renders *"Ask Codex to do anything… Powered by your ChatGPT account"* in a fresh profile with shared `~/.codex` auth).

### Actual

- SPA bundle loads fully (all ~250 assets HTTP 200), renders its startup loader.
- Over the webview port I observe the capnp bootstrap:
  ```
  webview → ["push",["pipeline",0,["services"]]]
  webview → ["pull",1]
  ext     → ["resolve",1,{customAvatars:{…}, fileAttachments:{…}}]
  ```
[...]

PR #805, opened 2026-06-04
fix(views): register a fallback default view container so contributed views don't crash registration

nikolai-vysotskyi:
### What
Register a minimal default **Sidebar** view container (`workbench.view.explorer`) inside the views service-override when none exists, so VS Code's `views` extension-point fallback resolves instead of throwing.

### Why
In a partial setup that doesn't include the files explorer, `ViewsExtensionHandler.addViews` resolves a contributed view's target container as `container ?? getDefaultViewContainer()`, where `getDefaultViewContainer()` is `viewContainersRegistry.get('workbench.view.explorer')`. With no Explorer registered that's `undefined`, and the next line reads `.extensionId` off it → an **uncaught** `TypeError`, which aborts the **entire** `contributes.views` pass. The net effect: *every* extension's contributed sidebar views silently fail to register (hit with `anthropic.claude-code` and `openai.chatgpt`, but it's generic — `ms-vscode.js-debug` contributing into the missing `debug` container triggers it too).

Stack:
```
TypeError: Cannot read properties of undefined (reading 'extensionId')
    at ViewsRegistry.addViews (Array.forEach)
    at …_handleExtensionPoint
    at …_resolveAndProcessExtensions
    at …_initialize
```

Full investigation + traces: #804.

### Change
`ensureDefaultViewContainer()` in `src/service-override/views.ts`, called at the top of `getServiceOverride()`. It registers a `hideIfEmpty` default Sidebar container **only if** `workbench.view.explorer` isn't already registered, so setups that DO include the explorer (or already register it) are unaffected. This mirrors the workaround discussed in #804.

### Notes
- An alternative would be a defensive null-guard in `viewsExtensionPoint.addViews` (a `vscode-patch`) so the fallback degrades without throwing even when no default container exists at all — happy to switch to that approach instead if you'd prefer it.
- Verified in our app: the repeated uncaught `extensionId` crash drops to **0** and contributed sidebar views register + render (a remote extension's sidebar webview-view now shows). I was **not** able to run the full repo build locally (it pulls the VS Code sources via `install-vscode`), so please rely on CI / a quick local check for the build.

Refs #804

### DIFF
--- a/src/service-override/viewCommon.ts
+++ b/src/service-override/viewCommon.ts
@@ -1,5 +1,12 @@
 import type { IEditorOverrideServices } from 'vs/editor/standalone/browser/standaloneServices'
 import { SyncDescriptor } from 'vs/platform/instantiation/common/descriptors'
+import { Registry } from 'vs/platform/registry/common/platform'
+import {
+  Extensions as ViewContainerExtensions,
+  type IViewContainersRegistry,
+  ViewContainerLocation
+} from 'vs/workbench/common/views'
+import { ViewPaneContainer } from 'vs/workbench/browser/parts/views/viewPaneContainer'
 import { IViewDescriptorService } from 'vs/workbench/common/views.service'
 import { ViewDescriptorService } from 'vs/workbench/services/views/browser/viewDescriptorService'
 import { IActivityService } from 'vs/workbench/services/activity/common/activity.service'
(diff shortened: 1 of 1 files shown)


CGNonofr:
Thanks! it would probably better to be in viewCommon though, so that it also impacts the workbench service override

src/service-override/views.ts
```diff
@@ -796,6 +802,41 @@ onRenderWorkbench(async (accessor) => {
...
+      title: { value: 'Explorer', original: 'Explorer' },
+      ctorDescriptor: new SyncDescriptor(ViewPaneContainer, [
+        EXPLORER_VIEW_CONTAINER_ID,
+        { mergeViewWithContainerWhenSingleView: true }
+      ]),
+      hideIfEmpty: true,
+      order: 0
+    },
+    ViewContainerLocation.Sidebar,
+    { isDefault: true }
```
CGNonofr: Does it really need to mimick the explorer settings, and any other value would be fine (except isDefault: true and the ViewContainerLocation)?

CGNonofr: Ok I've just checked the code, and it seems that `workbench.view.explorer` is required :)

nikolai-vysotskyi:
Moved it to `viewCommon.ts` as you suggested, so it also covers the workbench service override — and kept `workbench.view.explorer` as the id (thanks for confirming that's the one `getDefaultViewContainer()` requires). Force-pushed; the diff is now a single file.

On the null-guard alternative from #804 — I agree it'd be cleaner/less hacky, but it'd live as a `vscode-patch` on `viewsExtensionPoint`, and I couldn't run the full `install-vscode` build locally to generate/verify the patch. Happy to switch to that route if you prefer — could you point me at how you'd like the vscode patch regenerated, or take it from here if that's quicker? Otherwise this `viewCommon` version is ready to go.

CGNonofr:
Maybe https://github.com/CodinGame/monaco-vscode-api/wiki/How-to-upgrade-to-next-vscode-and-monaco%E2%80%90editor-versions can help you

nikolai-vysotskyi:
Thanks for the pointer! I tried to take the guard route, but I can't reproduce your `install-vscode` build environment here (no `mise`, and pulling/patching/building the full VS Code source to generate + verify a real `vscode-patch` isn't feasible on my side), so I can't ship a verified patch myself.

In case it's quick for you, here's exactly where it throws — in `viewsExtensionPoint.ts` (`ViewsExtensionHandler.addViews`), deminified from the bundle:

```js
const l = this.getViewContainer(key)            // undefined for an unknown container id
l || collector.warn(/* "… added to 'Explorer'." */)
const d = l ?? this.getDefaultViewContainer()   // getDefaultViewContainer() => viewContainersRegistry.get('workbench.view.explorer')
// …later, in the per-view loop:
… d.extensionId …                               // TypeError when d is undefined (no target container AND no Explorer)
```

So the guard would be: when `d` is `undefined` (neither the target container nor the Explorer default exists), `warn` + skip those views instead of dereferencing `d` — e.g. right after the `const d = …` line:

```js
if (d == null) {
  collector.warn(/* container '{key}' not found and no default view container is registered */)
  continue
}
```

Totally your call on which you prefer — the guard (you'd own the `vscode-patch`), or merging the `viewCommon` registration in #805 as-is (it's verified working on our side). Happy to adjust #805 either way.

CGNonofr:
What about https://github.com/CodinGame/monaco-vscode-api/pull/806 ?

nikolai-vysotskyi:
Superseded by #806 — CGNonofr's source-level guard in viewsExtensionPoint.addViews is the cleaner fix. Thanks!

PR #812, opened 2026-06-18
Demo persistent files

CGNonofr:
Just discovered that possibility, supported on all browsers :)


kaisalmen:
@CGNonofr commitlint does not like your firsst commit, otherwise this is good to go.

CGNonofr:
> @CGNonofr commitlint does not like your firsst commit, otherwise this is good to go.

Thanks! reworded

kaisalmen (approved):
LGTM

PR #813, opened 2026-06-24
Fix style conflict

CGNonofr:
By only injecting the styles in the shadow dom

Basically, when using the sandbox+shadow dom mode AND a regular editor in the same page, both vscode instances are conflicting on the styles

Currently, dynamic styles are injected into the document head, and then synchronized with the shadow dom.
Now, the dynamic styles are only injected in the shadow dom, except for font-faces that are extracted and injected into the page head, just like what is already done [here](https://github.com/CodinGame/monaco-vscode-api/blob/c88b9fb9dffcfb839da8a451c978f6d405436125/src/css.ts#L24-L40) for static stylesheets

### DIFF
--- a/vscode-patches/0060-feat-support-shadow-dom.patch
+++ b/vscode-patches/0060-feat-support-shadow-dom.patch
@@ -5,7 +5,7 @@ Subject: [PATCH] feat: support shadow dom
 
 ---
  src/vs/base/browser/dom.ts                    | 61 ++++++++++++++++---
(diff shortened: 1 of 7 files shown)


onora-hbl (dismissed):
LGTM but it doesn't compile

CGNonofr:
> LGTM but it doesn't compile

It was just an unused import, cleaned & amended

onora-hbl (approved):

PR #814, opened 2026-06-29
Use dom manipulation tools everywhere

CGNonofr:
Those tools supports shadow dom

The main consequence was the keybindings not working in shadow dom when the focus is in a webview (including the F1 - open quickaccess - command)


onora-hbl (approved):
LGTM

Issue #827, opened 2026-09-17
Selectively load ext-host `mainThread*` bridges so unused feature contributions (chat/notebook/mcp) can tree-shake out


brianjenkins94:
### Summary

`getExtensionServiceOverride()` unconditionally pulls in **every** extension-host `mainThread*` API bridge, and each bridge drags its feature's `contrib/**` into the bundle — even when the corresponding service override isn't installed. On a workbench that doesn't use chat/notebook/mcp, that's ~2.1 MB of source (~450–500 KB gzip) of dead code that can't currently be removed.

### Where it comes from

`src/service-override/extensions.ts` (line 72):

```ts
import 'vs/workbench/api/browser/extensionHost.contribution'
```

`extensionHost.contribution.ts` side-effect-imports ~90 `mainThread*` bridges (chat, notebook, mcp, speech, languageModel, …), each of which imports its feature's `contrib` services/UI. So enabling the extensions service — required to run any extension — pulls all of them in.

Confirmation that it isn't gated by the overrides: `src/service-override/{chat,notebook,mcp}.ts` import **no** `mainThread*` / `extensionHost` bridge, and the base (`src/contributions.ts`, `src/services.ts`) only carries a small curated contrib list plus DI-identifier re-exports. So there's currently no way to drop, say, chat by simply not calling `getChatServiceOverride()`.

### Measured impact (v36.2.7, minified consumer build)

Attributing a consumer `main.js` back to sources via its sourcemap (`sourcesContent`):

- `vs/workbench/contrib/chat` ≈ 1.03 MB source
- `vs/workbench/contrib/notebook` ≈ 1.15 MB source
- `vs/workbench/contrib/mcp` ≈ 38 KB source

All present despite chat/mcp being disabled and **no** notebook override installed. Coarse build-time stubbing of those `contrib` subtrees fails to build (`MISSING_EXPORT`), because the `*.service.js` identifier files are imported by name across the DI wiring — so only a proper source split works.

### Proposed change

Follow the existing split pattern (`vscode-patches/0037-refactor-split-workbench-contribution.patch`, `vscode-patches/0073-refactor-split-contribution.patch`):

1. Split `vs/workbench/api/browser/extensionHost.contribution.ts` — move feature-specific `mainThread*` imports into per-feature files (`extensionHost.chat.contribution.ts`, `.notebook.`, `.mcp.`, …), leaving only the universal bridges in the base aggregator.
[...]

PR #828, opened 2026-09-17
refactor: split extension-host contribution per feature

brianjenkins94:
Follows up on #827.

## Problem

`getExtensionServiceOverride()` pulls `vs/workbench/api/browser/extensionHost.contribution`, which unconditionally registers every `mainThread*` API bridge. Each bridge imports its feature's `contrib`, so a workbench that enables extensions but not chat/notebook/mcp still bundles the chat + notebook + mcp `contrib` it never uses. Omitting the service override doesn't help, because the bridges live in the always-on aggregator rather than the overrides.

## Change

Split the chat / language-model / AI / speech, notebook, and mcp `mainThread` bridges out of `extensionHost.contribution` into per-feature `extensionHost.<feature>.contribution` files (new vscode-patch `0109`), each imported by the matching service override — so a bridge, and the `contrib` it pulls, loads only when that override is installed. Mirrors the existing per-feature splits (patches `0037`, `0073`).

`mainThreadChatOutputRenderer` is left in the base: `mainThreadWebviewManager` instantiates it directly, so gating it would be a no-op.


CGNonofr:
Thanks, approach looks good!

I would have just preferred for the VSCode patch to not be breaking the logic: nothing imports your new `extensionHost.<feature>.contribution` files from the VSCode codebase, I would probably prefer having a extensionHost.base.contribution, and make extensionHost.contributions import extensionHost.base.contribution and all the `extensionHost.<feature>.contribution`

vscode-patches/0109-refactor-split-extension-host-contribution-per-feature.patch
```diff
@@ -86,42 +237,104 @@ index d42dedd..3e59ede 100644
...
++    MainContext.MainThreadChatSessions,
++    MainContext.MainThreadNotebook,
++    MainContext.MainThreadNotebookKernels,
++    MainContext.MainThreadNotebookDocuments,
++    MainContext.MainThreadNotebookEditors,
++    MainContext.MainThreadNotebookRenderers,
++    MainContext.MainThreadMcp
++]) {
++    extHostNamedCustomer(id)(NoopMainThreadCustomer);
++}
```
CGNonofr, line 338: Are you sure it supports well overriding an instance? because it seems to maintain a list of customer, so both noop and real instances will be in the list and instantiated 🤔

brianjenkins94: I'll downgrade this to a draft and check and see if this breaks any of my usage of the lib.

